PR #116, opened 2020-10-29
fix crash in head pose estimation demo

oscar-lima:
Boolean not parsed correctly to integer in some opencv releases (20.04 in particular)


oarriaga:
Thank you :+1: !

PR #125, opened 2020-11-16
ReadMe.md update for running on a remote machine 

hitman996:
Added instructions in the **_ReadMe.md_** file for training on a remote machine via ssh.


oarriaga (approved):
Thank you! Yes I think I will eventually delete the run_experiments.sh since every user should program their own.

PR #139, opened 2021-04-05
Fix read NoneType

nodece:
when the `self.step()` returns `None`, this code will crash, so this loop should be skipped.

camera often don't work in macOS Big Sur :(


oarriaga:
Hi :)

I checked it and you are right when an image `frame` is not detected in `step` it returns `None` but I am not accounting for that case in the `run` or `record` methods.

Thank you :)

PR #186, opened 2022-04-21
Merge minimal hand with paz library

proneetsharma:
- Transfer the example code to paz
- Load model weight and image from the web
- Update docs/structure.py


oarriaga:
Thank you! It looks good to me.

PR #213, opened 2022-08-01
Minimal hand detection

proneetsharma:
Add a pipeline that:
- Detection of minimal hand
- Estimate pose for the detected minimal hand
- Classify the hand closure


oarriaga:
Hi hi it seems like a single unit test broke. The test_match_box. Could you have a look at it? It most likely has to do with the changes done to the match function. Maybe the test was using the previous match (now match_beta) and now since we switched back to the default "match" it's not working anymore.

PR #232, opened 2022-11-03
Efficientdet

Manojkumarmuru:


examples/efficientdet/utils.py
```diff
@@ -50,12 +50,12 @@ def get_class_name_efficientdet(dataset_name):
                 'pottedplant', 'sheep', 'sofa', 'train', 'tvmonitor']
 
 
-class CustomDropout(keras.layers.Layer):
+class GetDropConnect(keras.layers.Layer):
```
oarriaga, line 53: I think we can remove the verb get from layer definition.

oarriaga (approved):

PR #294, opened 2023-07-04
Efficientdet

Manojkumarmuru:


Manojkumarmuru:
The following changes/updates have been made:

1. The bug with instantiating EfficientDet with `base_weights` and `head_weights` set to `None` is now resolved.
2. Parameters of AugmentDetection module inside paz.pipelines namely the `lower` of `RandomSaturation` and `probability` of `RandomSampleCrop` have been set to 0.7 and 1.0 respectively to achieve better mAP for SSD300 on VOC dataset (The maximum achieved mAP now is 76).
3. All (unused) modules that were present outside paz inside examples/object_detection have been removed.
4. All test cases pass.

PR #300, opened 2023-07-11
Add image synthesis examples

Jieying-Li:


Jieying-Li:
Add image synthesis examples

PR #332, opened 2024-01-18
Efficientpose

Manojkumarmuru:


paz/pipelines/pose.py
```diff
@@ -468,3 +472,297 @@ def _build_name_to_sizes(self):
...
+                                   [0, 1, 2, 3, 5]))
+        self.add(pr.ControlMap(self.preprocess_image, [0], [0, 1, 2]))
+        self.add(pr.ControlMap(self.scale_boxes, [3, 1], [3], keep={1: 1}))
+        self.add(pr.ControlMap(self.preprocess_boxes, [4], [5], keep={4: 4}))
+        self.add(pr.ControlMap(pr.TransformRotation(num_pose_dims), [3], [3]))
+        self.add(pr.ControlMap(self.match_poses, [4, 3], [3], keep={4: 4}))
+        self.add(pr.ControlMap(self.match_poses, [4, 5], [7], keep={4: 4}))
+        self.add(pr.ControlMap(self.concatenate_poses, [3, 8], [8],
+                               keep={3: 3}))
+        self.add(pr.ControlMap(self.concatenate_scale, [8, 1], [8]))
```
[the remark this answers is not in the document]
oarriaga: Rethink if the order of the arguments needs to be modified. Is hard because it's function dependent but also module dependent.

paz/pipelines/pose.py
```diff
@@ -468,3 +472,297 @@ def _build_name_to_sizes(self):
...
+            pr.SubtractMeanImage(mean),
+            pr.PadImage(model.input_shape[1]),
+            pr.CastImage(float),
+            pr.ExpandDims(axis=0)])
+        self.compute_camera_parameter = pr.ComputeCameraParameter(
+            camera_matrix, translation_scale_norm)
+
+    def call(self, image):
+        resizing_shape, image_scale = self.compute_resizing_shape(image)
+        resize_image = pr.ResizeImage(resizing_shape)
```
oarriaga, line 589: This is not a good idea since one is instantiating and destroying a class at every call. Rather, make a processor that takes as input the resizing shape. So in the constructor build resize_image = pr.ResizeImageWithShape() and then this could be applied in the call as resize_image(image, image_shape)

paz/pipelines/pose.py
```diff
@@ -468,3 +472,297 @@ def _build_name_to_sizes(self):
...
+            rotations = self.squeeze(rotations)
+            rotations = rotations[selected_indices]
+            rotations = self.transform_rotations(rotations)
+            translation_xy_Tz = self.regress_translation(translations)
+            translation = self.compute_tx_ty_tz(translation_xy_Tz,
+                                                camera_parameter)
+            translations = translation[selected_indices]
+        poses6D = self.to_pose_6D(box_data, rotations, translations)
+        return boxes2D, poses6D
+
```
oarriaga, line 655: Can we make this more compact?

paz/pipelines/pose.py
```diff
@@ -468,3 +472,297 @@ def _build_name_to_sizes(self):
...
+        postprocess: Callable.
+        draw_boxes2D: Callable.
+        wrap: Callable.
+
+    # Methods
+        _build_draw_pose6D()
+        call()
+    """
+    def __init__(self, model, class_names, score_thresh, nms_thresh,
+                 LINEMOD_CAMERA_MATRIX, LINEMOD_OBJECT_SIZES, preprocess=None,
```
oarriaga, line 697: Since this is a generic class one should not use LINEMOD at all

paz/pipelines/pose.py
```diff
@@ -468,3 +472,297 @@ def _build_name_to_sizes(self):
...
+            rotations = self.transform_rotations(rotations)
+            translation_xy_Tz = self.regress_translation(translations)
+            translation = self.compute_tx_ty_tz(translation_xy_Tz,
+                                                camera_parameter)
+            translations = translation[selected_indices]
+        poses6D = self.to_pose_6D(box_data, rotations, translations)
+        return boxes2D, poses6D
+
+
+class DetectAndEstimateEfficientPose(Processor):
```
oarriaga, line 657: Detect and Estimate seem to related. Can we just not call it something similar EstimateEfficientPose?

paz/pipelines/pose.py
```diff
@@ -468,3 +472,297 @@ def _build_name_to_sizes(self):
...
+            image = self.draw_boxes2D(image, boxes2D)
+        if self.show_poses6D:
+            self.draw_pose6D = self._build_draw_pose6D(
+                self.class_to_sizes, self.camera_matrix)
+            for box2D, pose6D in zip(boxes2D, poses6D):
+                image = self.draw_pose6D[box2D.class_name](image, pose6D)
+        return self.wrap(image, boxes2D, poses6D)
+
+
+class EfficientPosePhi0LinemodDriller(DetectAndEstimateEfficientPose):
```
oarriaga, line 744: this should be outsize.

paz/processors/image.py
oarriaga, line 1: check if I am using the probability like that.

paz/processors/standard.py
```diff
@@ -482,3 +482,14 @@ def __init__(self, topics):
...
+
+class ComputeSelectedIndices(Processor):
+    """Computes row-wise intersection between two given
+    arrays and returns the indices of the intersections.
+    """
+    def __init__(self):
+        super(ComputeSelectedIndices, self).__init__()
+
+    def call(self, box_data_raw, box_data):
+        return compute_selected_indices(box_data_raw, box_data)
```
oarriaga, line 495: We have to look if this is necessary? If so maybe a different name would also be better.

paz/processors/pose.py
```diff
@@ -127,3 +132,249 @@ def call(self, boxes2D):
             y_center = (z_center * v) / self.focal_length
             hands_center.append([x_center, y_center, z_center])
         return np.array(hands_center)
+
+
+class ComputeCameraParameter(Processor):
```
oarriaga, line 137: This name is misleading. It seems a rearrangment of the camera matrix?

paz/processors/pose.py
```diff
@@ -127,3 +132,249 @@ def call(self, boxes2D):
...
+        self.prior_boxes = prior_boxes
+        self.iou = iou
+        super(MatchPoses, self).__init__()
+
+    def call(self, boxes, poses):
+        return match_poses(boxes, poses, self.prior_boxes, self.iou)
+
+
+class TransformRotation(Processor):
+    """Computes axis angle rotation vector from a rotation matrix.
```
oarriaga, line 178: The name does not seem to correspond to the documentation? The name seems also very vague.

paz/processors/pose.py
```diff
@@ -127,3 +132,249 @@ def call(self, boxes2D):
...
+    """
+    def __init__(self, num_pose_dims):
+        self.num_pose_dims = num_pose_dims
+        super(TransformRotation, self).__init__()
+
+    def call(self, rotations):
+        return transform_rotation(rotations, self.num_pose_dims)
+
+
+class ConcatenatePoses(Processor):
```
oarriaga, line 195: Do we really need this as a processor? If it would not be extensivel used in other pipelines, we should maybe just call it as backend function in the call of the pipeline in which is being used.

oarriaga: I don't really know. Just something to think about.

paz/processors/pose.py
```diff
@@ -127,3 +132,249 @@ def call(self, boxes2D):
...
+            containing the transformed rotation.
+    """
+    def __init__(self):
+        super(ConcatenateScale, self).__init__()
+
+    def call(self, poses, scale):
+        return concatenate_scale(poses, scale)
+
+
+class Augment6DOF(Processor):
```
oarriaga, line 223: name should be different

oarriaga (changes requested):

paz/backend/poses.py
```diff
@@ -0,0 +1,191 @@
...
+        List: Containing transformation matrix, angle, scale
+    """
+    cx = LINEMOD_CAMERA_MATRIX[0, 2]
+    cy = LINEMOD_CAMERA_MATRIX[1, 2]
+    angle = np.random.uniform(angle_min, angle_max)
+    scale = np.random.uniform(scale_min, scale_max)
+    return [cv2.getRotationMatrix2D((cx, cy), -angle, scale), angle, scale]
+
+
+def compute_box_from_mask(mask, mask_value):
```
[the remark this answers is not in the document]
Manojkumarmuru: This is now moved to paz/backend/standard.py

paz/backend/image/image.py
```diff
@@ -351,3 +352,194 @@ def scale_resize(image, image_size):
...
+
+    # Arguments
+        image: Array, raw image.
+
+    # Returns:
+        contrasted: Array, contrast enhanced image.
+
+    # References:
+        [Python Pillow autocontrast](
+            https://github.com/python-pillow/Pillow/blob/main'
```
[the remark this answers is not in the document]
Manojkumarmuru: The paper uses autocontrast from PIL which has a default cutoff=0. Which theoretically is equivalent to histogram equalization. Since we have histogram equalization as a paz processor this Autocontrast computation is redundant. Hence this is now fully removed from paz.

paz/pipelines/pose.py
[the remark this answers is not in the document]
Manojkumarmuru: > I think this seems to complex. Can we change it for just a normal call function? Don't use control map so much.

In train.py `ProcessingSequence` requires a processor of class `SequentialProcessor` and does not work with class `Processor`, hence using call() inside `SequentialProcessor` class does not work.

paz/processors/pose.py
Manojkumarmuru: `ConcatenatePoses` is used in the pipeline `AugmentEfficientPose` which is inherited from the class `SequentialProcessor`.  Hence the pipeline `AugmentEfficientPose` does not have call() function.

paz/processors/pose.py
Manojkumarmuru: This processor is now removed because its only re-arrangement of camera_matrix and addition of image_scale and translation_scale_norm. These values are directly passed and used now wherever required instead of accessing from camera_parameter

paz/processors/image.py
Manojkumarmuru: paz/processors/geometric.py -> RandomRotation uses `probability` in a similar way.

PR #339, opened 2024-05-12
Updated unit tests

Manojkumarmuru:
The MiniXception model is instantiated programatically; only weights are loaded from the hdf5 file for unit tests to pass with Keras 2.0 and 3.0.


Manojkumarmuru:
All EfficientPose and its supporting modules have been added to the paz documentation.

PR #350, opened 2025-03-28
Refactor boxes.py and geometric.py to make them follow the expected convention

mohamedamrfaroukebrahim:


paz/backend/boxes.py
```diff
@@ -4,377 +4,552 @@
 import paz
 
 
-def split(boxes):
+def split_into_components(boxes):
```
oarriaga: I think is OK to leave it as "split" only. Since the API that we will be using is paz.boxes.split.

paz/backend/boxes.py
```diff
@@ -4,377 +4,552 @@
...
+    """Split boxes into x_min, y_min, x_max, y_max components."""
     return jp.split(boxes, 4, axis=1)
 
 
-def join(coordinate_0, coordinate_1, coordinate_2, coordinate_3):
-    return jp.concatenate(
-        [coordinate_0, coordinate_1, coordinate_2, coordinate_3], axis=1
-    )
+def get_center_coordinates(x_minimum, x_maximum, y_minimum, y_maximum):
```
oarriaga: We can most likely here call it "compute_centers". Since we are always expected to work with "coordinates" and we are "computing" them not "getting them". Moreover, we can here us x_min, y_min, x_max, y_max. We can use "max" and "min" as abbreviations.

paz/backend/boxes.py
```diff
@@ -4,377 +4,552 @@
...
+    """Compute center coordinates of boxes."""
+    center_x = (x_maximum + x_minimum) / 2.0
+    center_y = (y_maximum + y_minimum) / 2.0
+    return center_x, center_y
+
+
+def get_width_and_height(x_minimum, x_maximum, y_minimum, y_maximum):
```
oarriaga: Here we can call this "compute_size" or "compute_HW". Also we should return H first and then W. The min, max applies here as well.

paz/backend/boxes.py
```diff
@@ -4,377 +4,552 @@
...
+    """Decode center x-coordinate from predictions."""
+    return predictions[:, 0:1] * priors[:, 2:3] * variances[0] + priors[:, 0:1]
+
+
+def decode_center_y(predictions, priors, variances):
+    """Decode center y-coordinate from predictions."""
+    return predictions[:, 1:2] * priors[:, 3:4] * variances[1] + priors[:, 1:2]
+
+
+def decode_width(predictions, priors, variances):
```
oarriaga: Here we can call it decode_W

paz/backend/boxes.py
```diff
@@ -4,377 +4,552 @@
...
+    """Decode width from predictions."""
+    return priors[:, 2:3] * jp.exp(predictions[:, 2:3] * variances[2])
+
+
+def decode_height(predictions, priors, variances):
```
oarriaga: Here can call it decode_H

paz/backend/boxes.py
```diff
@@ -4,377 +4,552 @@
...
         paz.image.show(image, window_name, False)
         if key_is_pressed("q"):
             break
     cv2.destroyWindow(window_name)
-    return jp.array(boxes)
+
+
+def convert_to_center_form(matched_boxes):
+    """Convert matched boxes to center form."""
+    return to_center_form(matched_boxes[:, :4])
```
oarriaga: I think the outer function does not convey much more meaning that the inner function. Most likely we should call only the inner function when needed.

paz/backend/boxes.py
```diff
@@ -4,377 +4,552 @@
...
+    Returns:
+        array: Boxes in corner format [x_min, y_min, x_max, y_max]
     """
-    center_x, center_y, W, H = split(boxes)
-    x_min = center_x - (W / 2.0)
-    x_max = center_x + (W / 2.0)
-    y_min = center_y - (H / 2.0)
-    y_max = center_y + (H / 2.0)
-    return jp.concatenate([x_min, y_min, x_max, y_max], axis=1)
+    center_x, center_y, W, H = split_into_components(boxes)
```
oarriaga: we should call it "split"

paz/backend/boxes.py
```diff
@@ -4,377 +4,552 @@
...
 
-def from_selection(image, radius=5, color=(255, 0, 0), window_name="image"):
-    points, boxes = [], []
-    image = image.copy()
+def on_double_click_event(event, x, y, flags, param, image, points, boxes, radius, color):
+    """Handle double-click event to draw points and boxes."""
+    if event == cv2.EVENT_LBUTTONDBLCLK:
+        draw_circle_and_store_point(image, points, x, y, radius, color)
+        if len(points) % 2 == 0:
+            process_pair_of_points(image, points, boxes, color, radius)
```
oarriaga: This functions will most likely not be needed in the public API. Let's define inside the original function "from_selection" all the functions that are explicitly useful for this case. If we ever see that we might need them somewhere else we can put them out.

paz/backend/boxes.py
```diff
@@ -4,377 +4,552 @@
...
+def run_NMS_loop(condition_fn, step_fn, init_state):
+    """Run the NMS loop using lax.while_loop."""
+    return lax.while_loop(condition_fn, step_fn, init_state)
+
+
+def process_final_results(final_state, top_k):
+    """Process final NMS results to get selected indices."""
+    count, _, indices = final_state
+    num_selected = jp.minimum(count, top_k)
+    selected = lax.dynamic_slice(indices, (0,), (num_selected,))
```
oarriaga: I don't know if this function is jittable? Would it be possible to confirm this, maybe a unit test?

paz/processors/geometric.py
```diff
@@ -1,294 +1,505 @@
...
-    masked_boxes = masked_boxes.at[:, :2].set(
-        jp.maximum(masked_boxes[:, :2], image_crop_box[:2]) - image_crop_box[:2]
-    )
-    masked_boxes = masked_boxes.at[:, 2:].set(
-        jp.minimum(masked_boxes[:, 2:], image_crop_box[2:]) - image_crop_box[:2]
+    mask = (
+        (boxes[:, 0] < crop_region[2])
+        & (boxes[:, 1] < crop_region[3])
+        & (boxes[:, 2] > crop_region[0])
+        & (boxes[:, 3] > crop_region[1])
```
oarriaga: This is not so clear to me. I can't tell what coordinates we are comparing. Maybe we can split them before, give them name and then compare them.

oarriaga (changes requested):

paz/backend/boxes.py
```diff
@@ -5,60 +5,47 @@
...
-    """Transform from corner coordinates to center coordinates.
-
-    # Arguments
-        boxes: Numpy array with shape `(num_boxes, 4)`.
-
-    # Returns
-        Numpy array with shape `(num_boxes, 4)`.
-    """
-    x_min, y_min, x_max, y_max = split(boxes)
+def compute_centers(x_min, x_max, y_min, y_max):
```
oarriaga: I think this function should take boxes as input, not the coordinates given that we will be calling:
`paz.boxes.computer_center(boxes)`

paz/backend/boxes.py
```diff
@@ -5,60 +5,47 @@
...
+    return center_x, center_y
 
-def to_xywh(boxes):
-    return xyxy_to_xywh(boxes)
 
-
-def xyxy_to_xywh(boxes):
-    x_min, y_min = boxes[:, 0:1], boxes[:, 1:2]
-    x_max, y_max = boxes[:, 2:3], boxes[:, 3:4]
+def compute_size(x_min, x_max, y_min, y_max):
```
oarriaga: This function should take as input boxes not the coordinates. If you are using this function as coordinates, consider splitting the boxes first the applying this function. If you are using this function and the coordinates are scalar values. Then  , let's define the function inside the function that is using it.

paz/backend/boxes.py
```diff
@@ -150,231 +120,434 @@ def key_is_pressed(key="q", time=20):
...
+def calculate_intersection(inner_x_min, inner_y_min, inner_x_max, inner_y_max):
+    """Calculate intersection area between two boxes."""
     inner_w = jp.maximum((inner_x_max - inner_x_min), 0)
     inner_h = jp.maximum((inner_y_max - inner_y_min), 0)
-    intersection_area = inner_w * inner_h
-    # calculating the union
+    return inner_w * inner_h
+
+
+def calculate_union(
```
oarriaga: Most  of the time we say "compute" rather than "calculate" since compute is smaller.

paz/backend/boxes.py
```diff
@@ -150,231 +120,434 @@ def key_is_pressed(key="q", time=20):
...
+def get_inner_coordinates(x_min_A, y_min_A, x_max_A, y_max_A, x_min_B, y_min_B, x_max_B, y_max_B):
+    """Compute inner coordinates for intersection area."""
     inner_x_min = jp.maximum(x_min_B, x_min_A)
     inner_y_min = jp.maximum(y_min_B, y_min_A)
     inner_x_max = jp.minimum(x_max_B, x_max_A)
     inner_y_max = jp.minimum(y_max_B, y_max_A)
+    return inner_x_min, inner_y_min, inner_x_max, inner_y_max
+
+
+def calculate_intersection(inner_x_min, inner_y_min, inner_x_max, inner_y_max):
```
oarriaga: Most of the time we say "compute" rather than "calculate" since compute is smaller.

oarriaga (changes requested):
There are some minor changes.

oarriaga (approved):

oarriaga (approved):

PR #354, opened 2025-08-18
Per face buffers

Priyageorge123:
Each detected face gets its own VVAD classifier (and buffer).


Priyageorge123:
Pushed to the wrong repo

PR #364, opened 2026-06-24
Add IMDB two-class face attribute classifier

oarriaga:
Ports the legacy `face_classification` "gender" capability under a neutral,
dataset-named API (`IMDB`), **with the original pretrained weights extracted
and hosted** — so it's verifiable end-to-end.

## What's added
- **`build_mini_xception_imdb`** — faithfully reproduces the original
  `oarriaga/face_classification` mini_XCEPTION (8-kernel stem, strided fourth
  module, `64x64x1`) so the released weights load by topological order. A
  leading `Rescaling` maps the `[0,1]` classifier input to the `[-1,1]`
  training range.
- **`MiniXceptionIMDB`** — loads the converted weights from `altamira-data`
  **v0.22**.
- **`ClassifyMiniXceptionIMDB` / `DetectMiniXceptionIMDB`** apps, factored over
  shared `ClassifyMiniXception` / `DetectMiniXception` helpers that the FER
  siblings now reuse (refactor, no duplication).
- **`labels('IMDB') -> ['man','woman']`** (the dataset's own labels; the rename
  was the public API).
- **`examples/imdb_classifier/`** — loader-free `train.py` (`.npy` arrays at
  `64x64`) and a `demo.py` that runs off a local checkpoint via the shared
  helpers.

## Weights provenance
The pretrained weights are `gender_mini_XCEPTION.21-0.95.hdf5` from
`oarriaga/face_classification` (Keras 2.0.5), converted to Keras-3 as
`imdb_mini_XCEPTION_paz_jax.weights.h5`. Conversion is **bit-identical**
(round-trip max abs diff `0.0`).

## Verification (CPU)
- Converted weights load by topological order; `52,658` params (matches source).
- **Fresh download + run**: cache cleared, `MiniXceptionIMDB()` downloads from
  v0.22, loads, and classifies real detected faces (confident on clear faces).
- `ClassifyMiniXceptionIMDB` / `DetectMiniXceptionIMDB` factories resolve and
  run end-to-end.
- **FER path intact** through the shared-helper refactor.
- Synthetic train smoke for `train.py` (loss decreases, weights file produced).

### DIFF
--- a/paz/models/classification/xception.py
+++ b/paz/models/classification/xception.py
@@ -301,21 +301,21 @@ def build_mini_xception_imdb(input_shape, num_classes, l2_reg=0.01):
     # by topological order. Rescaling maps the [0, 1] classifier input to the
     # [-1, 1] training range.
     reg = l2(l2_reg)
+    conv = dict(use_bias=False, kernel_regularizer=reg)
+    down = dict(strides=2, padding="same", use_bias=False)
+    sep = dict(padding="same", use_bias=False, depthwise_regularizer=reg)
     image = Input(input_shape)
     x = Rescaling(scale=2.0, offset=-1.0)(image)
-    x = Conv2D(8, 3, kernel_regularizer=reg, use_bias=False)(x)
+    x = Conv2D(8, 3, **conv)(x)
     x = Activation("relu")(BatchNormalization()(x))
-    x = Conv2D(8, 3, kernel_regularizer=reg, use_bias=False)(x)
+    x = Conv2D(8, 3, **conv)(x)
     x = Activation("relu")(BatchNormalization()(x))
     for num_kernels in [16, 32, 64, 128]:
-        residual = Conv2D(num_kernels, 1, strides=2, padding="same",
-                          use_bias=False)(x)
+        residual = Conv2D(num_kernels, 1, **down)(x)
         residual = BatchNormalization()(residual)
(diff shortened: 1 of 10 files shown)


paz/models/classification/xception.py
```diff
@@ -293,3 +293,51 @@ def MiniXceptionFER():
...
+    x = Activation("relu")(BatchNormalization()(x))
+    for num_kernels in [16, 32, 64, 128]:
+        residual = Conv2D(num_kernels, 1, strides=2, padding="same",
+                          use_bias=False)(x)
+        residual = BatchNormalization()(residual)
+        x = SeparableConv2D(num_kernels, 3, padding="same",
+                            depthwise_regularizer=reg, use_bias=False)(x)
+        x = Activation("relu")(BatchNormalization()(x))
+        x = SeparableConv2D(num_kernels, 3, padding="same",
+                            depthwise_regularizer=reg, use_bias=False)(x)
```
oarriaga: Let's try to have to always have a single  line so it's easier to read. We could try to make either a helper function for that layer that uses repeated terms or we could also make some kwargs that are passed multiple times.

oarriaga: Done in 5e1afdc2 — packed the repeated layer args into shared kwargs dicts (`conv`/`down`/`sep`) so every Conv2D/SeparableConv2D call is one line. Pure refactor: params unchanged (52,658) and real-face scores identical.

PR #365, opened 2026-06-24
Port implicit orientation learning (augmented autoencoder)

oarriaga:
Faithful JAX / Keras-3 port of the legacy augmented-autoencoder (AAE) baseline
for implicit 3D orientation estimation. Synthetic training data is generated
with the built-in `paz.graphics` renderer — no external rasterizer.

## What's added
- **`paz.models.AutoEncoder((128,128,3), latent_dimension)`** — conv encoder →
  latent → conv decoder (sigmoid). `paz.models.extract_encoder(model)` slices
  the encoder for the codebook. Co-located test.
- **`examples/implicit_orientation_learning/`**:
  - `scenes.py` — mesh loading/normalization, sphere pose sampling, and view
    rendering via `paz.graphics` (colored-cube fallback so it runs offline).
  - `train.py` — renders clean views once and **caches them** (`views.npz`),
    then trains the AE with per-epoch domain randomization (background
    composite via the constant-background mask, occlusion, brightness).
  - `codebook.py` — builds a pose codebook (encode rendered grid) and does
    cosine nearest-neighbor retrieval.
  - `demo.py` — webcam → encode → nearest codebook view (loads a local
    checkpoint; no weights hosted).

## Design notes
- Ray tracing is the slow step, so views are rendered **once and cached**;
  training only augments the cache each epoch. `build_mesh` normalizes any mesh
  to unit extent, so the default camera distance frames any object.
- The decoder is only used in training; inference is encoder + codebook.

## Verification (CPU)
- Model build + encoder slice — `paz/models/autoencoder/autoencoder_test.py`
  passes (recon `(B,128,128,3)` in `[0,1]`, latent `(B,d)`).
- Renderer produces orientation-distinct views (foreground framing tuned).
- Synthetic train smoke: loss decreases; `aae.weights.h5` + cache produced.
- Codebook + cosine retrieval returns the correct pose.

## Scope
No weights are hosted (would need a specific object model), so the demo loads a
locally trained checkpoint — same pattern as the segmentation example.

### DIFF
--- a/examples/implicit_orientation_learning/codebook.py
+++ b/examples/implicit_orientation_learning/codebook.py
@@ -1,0 +1,26 @@
+from collections import namedtuple
+
+import numpy as np
+
+import scenes
+
+Codebook = namedtuple("Codebook", ["views", "poses", "latents"])
+
+
+def build_codebook(encoder, render_fn, poses):
+    views = scenes.render_views(render_fn, poses)
+    latents = np.asarray(encoder.predict(views, verbose=0))
+    return Codebook(views, np.stack([np.asarray(p) for p in poses]),
+                    unit_rows(latents))
+
+
+def closest_view(encoder, image, codebook):
+    latent = np.asarray(encoder(image[None]))[0]
+    similarities = codebook.latents @ unit_rows(latent[None])[0]
+    closest = int(np.argmax(similarities))
+    return codebook.views[closest], codebook.poses[closest]
+
+
+def unit_rows(vectors):
+    norms = np.linalg.norm(vectors, axis=-1, keepdims=True)
+    return vectors / np.maximum(norms, 1e-8)
--- a/examples/implicit_orientation_learning/codebook.py
+++ b/examples/implicit_orientation_learning/codebook.py
@@ -1,6 +1,6 @@
 from collections import namedtuple
 
-import numpy as np
+import jax.numpy as jp
 
 import scenes
 
@@ -9,19 +9,19 @@ Codebook = namedtuple("Codebook", ["views", "poses", "latents"])
 
 def build_codebook(encoder, render_fn, poses):
     views = scenes.render_views(render_fn, poses)
-    images = views.astype("float32") / 255.0
-    latents = np.asarray(encoder.predict(images, verbose=0))
-    return Codebook(views, np.stack([np.asarray(p) for p in poses]),
-                    unit_rows(latents))
+    images = jp.asarray(views, jp.float32) / 255.0
+    latents = unit_rows(jp.asarray(encoder.predict(images, verbose=0)))
+    matrices = jp.stack([jp.asarray(pose) for pose in poses])
+    return Codebook(jp.asarray(views), matrices, latents)
 
 
 def closest_view(encoder, image, codebook):
-    latent = np.asarray(encoder(image[None]))[0]
-    similarities = codebook.latents @ unit_rows(latent[None])[0]
-    closest = int(np.argmax(similarities))
+    latent = unit_rows(jp.asarray(encoder(image[jp.newaxis])))[0]
+    similarities = codebook.latents @ latent
+    closest = jp.argmax(similarities)
     return codebook.views[closest], codebook.poses[closest]
 
 
 def unit_rows(vectors):
-    norms = np.linalg.norm(vectors, axis=-1, keepdims=True)
-    return vectors / np.maximum(norms, 1e-8)
+    norms = jp.linalg.norm(vectors, axis=-1, keepdims=True)
+    return vectors / jp.maximum(norms, 1e-8)
--- a/examples/implicit_orientation_learning/codebook.py
+++ b/examples/implicit_orientation_learning/codebook.py
@@ -9,7 +9,8 @@ Codebook = namedtuple("Codebook", ["views", "poses", "latents"])
 
 def build_codebook(encoder, render_fn, poses):
     views = scenes.render_views(render_fn, poses)
-    latents = np.asarray(encoder.predict(views, verbose=0))
+    images = views.astype("float32") / 255.0
+    latents = np.asarray(encoder.predict(images, verbose=0))
     return Codebook(views, np.stack([np.asarray(p) for p in poses]),
                     unit_rows(latents))
 
--- a/examples/implicit_orientation_learning/demo.py
+++ b/examples/implicit_orientation_learning/demo.py
@@ -3,50 +3,38 @@ import argparse
 
 os.environ["KERAS_BACKEND"] = "jax"
 
-import numpy as np
-
 import paz
 import scenes
 import codebook
 
 WEIGHTS = "experiments/aae.weights.h5"
 
+parser = argparse.ArgumentParser(description="Implicit orientation demo")
+parser.add_argument("--weights", default=WEIGHTS)
+parser.add_argument("--mesh", default=None)
+parser.add_argument("--image_size", default=128, type=int)
+parser.add_argument("--distance", default=0.9, type=float)
+parser.add_argument("--camera", default=0, type=int)
+parser.add_argument("--H", default=480, type=int)
+parser.add_argument("--W", default=640, type=int)
+args = parser.parse_args()
 
-def ImplicitRotationPredictor(weights, mesh_path, image_size, distance):
-    model = paz.models.AutoEncoder((image_size, image_size, 3))
-    model.load_weights(weights)
-    encoder = paz.models.extract_encoder(model)
-    mesh = scenes.build_mesh(mesh_path)
-    render = scenes.build_renderer(mesh, image_size, distance)
-    poses = scenes.grid_poses(10, 10, distance)
-    book = codebook.build_codebook(encoder, render, poses)
-    size = (image_size, image_size)
-
-    def call(image):
-        crop = paz.image.resize_opencv(image, size)
-        crop = paz.image.normalize(np.asarray(crop, "float32"))
-        view, pose = codebook.closest_view(encoder, crop, book)
-        return paz.image.resize_opencv(np.asarray(view), size)
-
-    return call
-
-
-def main():
-    parser = argparse.ArgumentParser(description="Implicit orientation demo")
-    parser.add_argument("--weights", default=WEIGHTS)
-    parser.add_argument("--mesh", default=None)
-    parser.add_argument("--image_size", default=128, type=int)
-    parser.add_argument("--distance", default=0.9, type=float)
-    parser.add_argument("--camera", default=0, type=int)
-    parser.add_argument("--H", default=480, type=int)
-    parser.add_argument("--W", default=640, type=int)
-    args = parser.parse_args()
-    predict = ImplicitRotationPredictor(
-        args.weights, args.mesh, args.image_size, args.distance)
-    camera = paz.Camera(args.camera)
-    player = paz.VideoPlayer((args.H, args.W), predict, camera)
-    player.run()
+size = (args.image_size, args.image_size)
+model = paz.models.AutoEncoder((args.image_size, args.image_size, 3))
+model.load_weights(args.weights)
+encoder = paz.models.extract_encoder(model)
+mesh = scenes.build_mesh(args.mesh)
+render = scenes.build_renderer(mesh, args.image_size, args.distance)
+book = codebook.build_codebook(encoder, render,
+                               scenes.grid_poses(10, 10, args.distance))
 
 
-if __name__ == "__main__":
-    main()
+def predict(image):
+    crop = paz.image.normalize(paz.image.resize_opencv(image, size))
+    view, pose = codebook.closest_view(encoder, crop, book)
+    return paz.image.resize_opencv(view, size)
+
+
+camera = paz.Camera(args.camera)
+player = paz.VideoPlayer((args.H, args.W), predict, camera)
+player.run()
--- a/examples/implicit_orientation_learning/demo.py
+++ b/examples/implicit_orientation_learning/demo.py
@@ -9,32 +9,32 @@ import codebook
 
 WEIGHTS = "experiments/aae.weights.h5"
 
-parser = argparse.ArgumentParser(description="Implicit orientation demo")
-parser.add_argument("--weights", default=WEIGHTS)
-parser.add_argument("--mesh", default=None)
-parser.add_argument("--image_size", default=128, type=int)
-parser.add_argument("--distance", default=0.9, type=float)
-parser.add_argument("--camera", default=0, type=int)
-parser.add_argument("--H", default=480, type=int)
-parser.add_argument("--W", default=640, type=int)
-args = parser.parse_args()
 
-size = (args.image_size, args.image_size)
-model = paz.models.AutoEncoder((args.image_size, args.image_size, 3))
-model.load_weights(args.weights)
-encoder = paz.models.extract_encoder(model)
-mesh = scenes.build_mesh(args.mesh)
-render = scenes.build_renderer(mesh, args.image_size, args.distance)
-book = codebook.build_codebook(encoder, render,
-                               scenes.grid_poses(10, 10, args.distance))
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description="Implicit orientation demo")
+    parser.add_argument("--weights", default=WEIGHTS)
+    parser.add_argument("--mesh", default=None)
+    parser.add_argument("--image_size", default=128, type=int)
+    parser.add_argument("--distance", default=0.9, type=float)
+    parser.add_argument("--camera", default=0, type=int)
+    parser.add_argument("--H", default=480, type=int)
+    parser.add_argument("--W", default=640, type=int)
+    args = parser.parse_args()
 
+    size = (args.image_size, args.image_size)
+    model = paz.models.AutoEncoder((args.image_size, args.image_size, 3))
+    model.load_weights(args.weights)
+    encoder = paz.models.extract_encoder(model)
+    mesh = scenes.build_mesh(args.mesh)
+    render = scenes.build_renderer(mesh, args.image_size, args.distance)
+    poses = scenes.grid_poses(10, 10, args.distance)
+    book = codebook.build_codebook(encoder, render, poses)
 
-def predict(image):
-    crop = paz.image.normalize(paz.image.resize_opencv(image, size))
-    view, pose = codebook.closest_view(encoder, crop, book)
-    return paz.image.resize_opencv(view, size)
+    def predict(image):
+        crop = paz.image.normalize(paz.image.resize_opencv(image, size))
+        view, pose = codebook.closest_view(encoder, crop, book)
+        return paz.image.resize_opencv(view, size)
 
-
-camera = paz.Camera(args.camera)
-player = paz.VideoPlayer((args.H, args.W), predict, camera)
-player.run()
+    camera = paz.Camera(args.camera)
+    player = paz.VideoPlayer((args.H, args.W), predict, camera)
+    player.run()
--- a/examples/implicit_orientation_learning/demo.py
+++ b/examples/implicit_orientation_learning/demo.py
@@ -12,7 +12,26 @@ import codebook
 WEIGHTS = "experiments/aae.weights.h5"
 
 
-def parse_arguments():
+def ImplicitRotationPredictor(weights, mesh_path, image_size, distance):
+    model = paz.models.AutoEncoder((image_size, image_size, 3))
+    model.load_weights(weights)
+    encoder = paz.models.extract_encoder(model)
+    mesh = scenes.build_mesh(mesh_path)
+    render = scenes.build_renderer(mesh, image_size, distance)
+    poses = scenes.grid_poses(10, 10, distance)
+    book = codebook.build_codebook(encoder, render, poses)
+    size = (image_size, image_size)
+
+    def call(image):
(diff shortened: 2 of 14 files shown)


examples/implicit_orientation_learning/codebook.py
```diff
@@ -0,0 +1,27 @@
+from collections import namedtuple
+
+import numpy as np
+
+import scenes
+
+Codebook = namedtuple("Codebook", ["views", "poses", "latents"])
+
+
+def build_codebook(encoder, render_fn, poses):
```
oarriaga, line 10: This function and the whole codebook should be in jax

examples/implicit_orientation_learning/demo.py
```diff
@@ -0,0 +1,55 @@
...
+import numpy as np
+
+import paz
+import scenes
+import codebook
+
+WEIGHTS = "experiments/aae.weights.h5"
+
+
+def parse_arguments():
```
oarriaga: Inline the arguments in the main always (no helper function for argparse).

examples/implicit_orientation_learning/demo.py
```diff
@@ -0,0 +1,55 @@
...
+    parser.add_argument("--mesh", default=None)
+    parser.add_argument("--image_size", default=128, type=int)
+    parser.add_argument("--distance", default=0.9, type=float)
+    parser.add_argument("--camera", default=0, type=int)
+    parser.add_argument("--H", default=480, type=int)
+    parser.add_argument("--W", default=640, type=int)
+    return parser.parse_args()
+
+
+def build_pipeline(args):
```
oarriaga: This should have the same naming convention as an application module thus it should use CamelCase.

examples/implicit_orientation_learning/scenes.py
```diff
@@ -0,0 +1,72 @@
...
+def view_pose(theta, phi, distance):
+    x = distance * np.sin(theta) * np.cos(phi)
+    y = distance * np.cos(theta)
+    z = distance * np.sin(theta) * np.sin(phi)
+    origin = jp.array([x, y, z])
+    up = jp.array([0.0, 1.0, 0.0])
+    return paz.SE3.view_transform(origin, jp.zeros(3), up)
+
+
+def random_poses(num_views, distance, top_only, seed):
```
oarriaga: Can we not use paz.lie.SE3 or paz.lie.SO3 to sample this instead?

examples/implicit_orientation_learning/scenes.py
```diff
@@ -0,0 +1,72 @@
+import numpy as np
```
oarriaga, line 1: Just jax as much as possible specially where performance matters. Use jax.jit as well

examples/implicit_orientation_learning/train.py
```diff
@@ -0,0 +1,99 @@
...
+    mesh = scenes.build_mesh(mesh_path)
+    render = scenes.build_renderer(mesh, image_size, distance)
+    poses = scenes.random_poses(num_views, distance, False, seed)
+    views = scenes.render_views(render, poses)
+    masks = (views < 230).any(axis=-1)
+    np.savez(cache, views=views, masks=masks)
+    return views, masks
+
+
+def parse_arguments():
```
oarriaga: Same as before.

examples/implicit_orientation_learning/train.py
```diff
@@ -0,0 +1,99 @@
+import os
+
+os.environ["KERAS_BACKEND"] = "jax"
+
+import argparse
+
+import numpy as np
```
oarriaga, line 8: Use Jax instead as much as possible and jit when it does not trigger recompilation.

examples/implicit_orientation_learning/train.py
```diff
@@ -0,0 +1,99 @@
...
+        randomize = paz.image.randomize_rendered_image
+        image = randomize(key, view, mask, self.backgrounds)
+        return np.asarray(image, "float32") / 255.0
+
+
+def load_backgrounds(path, image_size):
+    if path is None:
+        return None
+    files = [os.path.join(path, f) for f in os.listdir(path)]
+    crops = [paz.image.resize_opencv(paz.image.load(f), (image_size,) * 2)
```
oarriaga, line 43: Make this a proper for loop (it does not fit as a single line) and use full English names in the iteration variables.

paz/backend/image.py
```diff
@@ -403,6 +403,80 @@ def random_color_transform(key, image):
     return image
 
 
+def make_random_plain_image(key, shape):
+    """Plain image of a single random RGB color."""
+    color = jax.random.randint(key, (shape[-1],), 0, 256)
+    return jp.broadcast_to(color, shape).astype(jp.uint8)
+
+
+def random_crop(key, image, size):
```
oarriaga: I believe we have code in paz that addresses this issue. Make sure the function can be jitted without triggering recompilation.

paz/backend/image.py
```diff
@@ -403,6 +403,80 @@ def random_color_transform(key, image):
...
+    if mask.ndim == 2:
+        mask = mask[..., jp.newaxis]
+    foreground = paz.cast(image, jp.float32)
+    background = paz.cast(background, jp.float32)
+    blended = mask * foreground + (1.0 - mask) * background
+    return paz.cast(blended, jp.uint8)
+
+
+def add_occlusion(key, image, max_radius_scale=0.5):
+    """Draws one random filled polygon over the image."""
```
oarriaga: We should build this in jax in a way that it does not trigger recompilation, and also build as a standalone the CV function in Jax. Test against it.

paz/backend/image.py
```diff
@@ -403,6 +403,80 @@ def random_color_transform(key, image):
...
+    angles = np.sort(np.asarray(jax.random.uniform(keys[3], (num_vertices,))))
+    offsets = np.stack([np.cos(angles * 2.0 * np.pi),
+                        np.sin(angles * 2.0 * np.pi)], axis=1)
+    points = (center + (radii * max_radius)[:, np.newaxis] * offsets)
+    color = np.asarray(jax.random.randint(keys[4], (3,), 0, 256)).tolist()
+    cv2.fillPoly(image, [points.astype(np.int32)], color)
+    return image
+
+
+def random_blur(key, image, kernel_size=9):
```
oarriaga, line 408: We should implement this in jax. Ultimately we should see what is the speed up in the Domain randomization with Jax and without Jax so that we can validate that using jax makes sense.

examples/implicit_orientation_learning/codebook.py
oarriaga: Done in e5843775 — `codebook.py` is now JAX: latents and cosine-similarity/argmax retrieval use `jax.numpy`.

examples/implicit_orientation_learning/demo.py
oarriaga: Done — argparse inlined in `main()`, no helper function.

examples/implicit_orientation_learning/demo.py
oarriaga: Done — renamed to a CamelCase application factory `ImplicitRotationPredictor(...)`.

examples/implicit_orientation_learning/scenes.py
oarriaga: Done — `scenes.py` now samples camera poses with `paz.SO3.sample` (random views) and `paz.SO3.rotation_x/y` (codebook grid); no manual trigonometry.

examples/implicit_orientation_learning/scenes.py
oarriaga: Done — `scenes.py` is fully JAX; the renderer (`mesh_renderer`) is already jitted and pose construction is jax.

examples/implicit_orientation_learning/train.py
oarriaga: Done — argparse inlined in `train.py:main()` too.

examples/implicit_orientation_learning/train.py
oarriaga: Done — the per-batch randomization is now `jax.jit(jax.vmap(...))`; verified it compiles once and does not recompile across batches.

examples/implicit_orientation_learning/train.py
oarriaga: Resolved by removing the loop entirely: the comprehension is replaced with a jitted `vmap` over the batch (per the JAX request above), so there is no per-sample Python iteration left.

paz/backend/image.py
oarriaga: Made `random_crop` jittable via `lax.dynamic_slice` (no Python int slicing), and `make_random_plain_image` is pure jax with a static shape. If you meant a specific existing paz helper I should reuse here instead, point me to it and I'll swap.

paz/backend/image.py
oarriaga: Done — added a standalone JAX `fill_polygon` (even-odd rasterizer); `add_occlusion` uses it with a fixed vertex count so it jits. Test `test_fill_polygon_matches_cv2` asserts >95% pixel agreement with `cv2.fillPoly`.

paz/backend/image.py
oarriaga: Done — added JAX `gaussian_blur` (depthwise conv; test matches cv2 interior). Benchmark (CPU, batch 32 @ 96x96): JAX jit+vmap ~30ms vs cv2 per-image ~1.7ms — cv2 wins on CPU, but the JAX path is jit/vmap-able, differentiable, and on-device with the renderer, so it should pay off on GPU. Kept JAX for the JAX-first lib; happy to revert to cv2 for occlusion/blur if you'd rather optimize CPU data-gen.

examples/implicit_orientation_learning/demo.py
```diff
@@ -0,0 +1,52 @@
...
+    parser.add_argument("--W", default=640, type=int)
+    args = parser.parse_args()
+    predict = ImplicitRotationPredictor(
+        args.weights, args.mesh, args.image_size, args.distance)
+    camera = paz.Camera(args.camera)
+    player = paz.VideoPlayer((args.H, args.W), predict, camera)
+    player.run()
+
+
+if __name__ == "__main__":
```
oarriaga: Let's not do this. We should inline the function here directly.

examples/implicit_orientation_learning/demo.py
```diff
@@ -0,0 +1,52 @@
+import os
+import argparse
+
+os.environ["KERAS_BACKEND"] = "jax"
+
+import numpy as np
```
oarriaga: Are we still using this? Always remove dead code.

examples/implicit_orientation_learning/scenes.py
```diff
@@ -0,0 +1,72 @@
...
+
+def camera_pose(rotation, distance):
+    origin = rotation @ jp.array([0.0, 0.0, distance])
+    world_up = rotation @ jp.array([0.0, 1.0, 0.0])
+    return paz.SE3.view_transform(origin, jp.zeros(3), world_up)
+
+
+def random_poses(key, num_views, distance):
+    keys = jax.random.split(key, num_views)
+    return [camera_pose(paz.SO3.sample(view_key), distance)
```
oarriaga: It does not fit in a single line, thus just make a real for loop.

examples/implicit_orientation_learning/train.py
```diff
@@ -0,0 +1,91 @@
+import os
+
+os.environ["KERAS_BACKEND"] = "jax"
+
+import argparse
+from functools import partial
+
+import numpy as np
```
oarriaga, line 8: Are we still using this? Remove all dead code.

examples/implicit_orientation_learning/train.py
```diff
@@ -0,0 +1,91 @@
...
+    callbacks = [
+        keras.callbacks.CSVLogger(os.path.join(args.root, "log.csv")),
+        keras.callbacks.ModelCheckpoint(weights, save_weights_only=True),
+        keras.callbacks.ReduceLROnPlateau("loss", patience=10),
+    ]
+    model.fit(sequence, epochs=args.epochs, callbacks=callbacks)
+
+
+if __name__ == "__main__":
+    main()
```
oarriaga: Always inline the whole main function here. No one single main function.

paz/backend/image.py
```diff
@@ -403,6 +403,114 @@ def random_color_transform(key, image):
...
+def fill_polygon(image, polygon, color):
+    """Fills a polygon in an image (JAX replacement for cv2.fillPoly)."""
+    grid_y, grid_x = jp.meshgrid(jp.arange(image.shape[0]),
+                                 jp.arange(image.shape[1]), indexing="ij")
+    inside = points_in_polygon(grid_x, grid_y, polygon)
+    color = paz.cast(color, image.dtype)
+    return jp.where(inside[..., jp.newaxis], color, image)
+
+
+def gaussian_blur(image, kernel_size=9, sigma=2.0):
```
oarriaga: I think we might have a function for this already in paz.backend

paz/backend/image.py
oarriaga: Let's look in paz.backend

paz/backend/image.py
oarriaga: You're right — reused the existing `paz.backend.image.apply_gaussian_blur` (generalized it to multi-channel via depthwise conv) and removed my duplicate `gaussian_blur`. `random_blur` now calls it. (Done in e0909870.)

paz/backend/image.py
oarriaga: Looked: the existing crops in paz.backend (`scene.random_crop`, `processors/geometric`) are detection-specific (crop + adjust boxes/IOU), not a plain fixed-window crop. Since backgrounds are pre-resized to the image size, no crop is actually needed — I dropped `random_crop` and `sample_background` just indexes a background. If you'd prefer reusing a specific helper for variety, point me to it.

examples/implicit_orientation_learning/demo.py
oarriaga: Done — `demo.py` is now a flat module-level script: no `main()`, no factory; argparse and the pipeline are inlined at module level.

examples/implicit_orientation_learning/demo.py
oarriaga: Removed — `numpy` was unused in `demo.py` (resize_opencv/normalize take and return jax arrays). pyflakes confirms no unused imports.

examples/implicit_orientation_learning/scenes.py
oarriaga: Done — `random_poses` is a real for loop now.

examples/implicit_orientation_learning/train.py
oarriaga: `numpy` is still used in `train.py` (the `.npz` view cache via np.load/savez and the host-array conversion keras' PyDataset returns). pyflakes confirms no unused imports remain.

examples/implicit_orientation_learning/train.py
oarriaga: Done — `train.py` is flat module-level now: no `main()`, argparse inlined; the data helpers (render_dataset, load_backgrounds, OrientationSequence) remain as functions.

paz/backend/image.py
oarriaga: GPU benchmark done (RTX A4000), 128x128 — the JAX path wins on GPU:

| pipeline | batch 32 | batch 128 |
|---|---|---|
| JAX jit+vmap, GPU | 0.7 ms | 2.2 ms |
| JAX jit+vmap, CPU | 47.8 ms | 191.7 ms |
| cv2 per-image, CPU | 6.9 ms | 25.8 ms |

So on GPU it's ~10-12x faster than cv2-on-CPU and ~70-87x faster than JAX-on-CPU; the renderer's views also stay on-device. Keeping the JAX version.

examples/implicit_orientation_learning/demo.py
```diff
@@ -0,0 +1,40 @@
+import os
+import argparse
+
+os.environ["KERAS_BACKEND"] = "jax"
+
+import paz
+import scenes
+import codebook
+
```
oarriaga, line 9: Do make the classic main function for python __name__ etc but don't abstract it with another main function in it, just inline the main function there. Apply that for all main functions in the PR.

paz/backend/image.py
```diff
@@ -403,6 +399,86 @@ def random_color_transform(key, image):
     return image
 
 
+def make_random_plain_image(key, shape):
+    """Plain image of a single random RGB color."""
+    color = jax.random.randint(key, (shape[-1],), 0, 256)
+    return jp.broadcast_to(color, shape).astype(jp.uint8)
+
+
+def blend_background(image, background, mask):
```
oarriaga: We might have already a function in paz for this. Look at all the backends.

paz/backend/image.py
```diff
@@ -403,6 +399,86 @@ def random_color_transform(key, image):
...
+    next_x, next_y = jp.roll(vertex_x, -1), jp.roll(vertex_y, -1)
+    x = grid_x[..., jp.newaxis]
+    y = grid_y[..., jp.newaxis]
+    straddles = (vertex_y > y) != (next_y > y)
+    slope = (next_x - vertex_x) * (y - vertex_y) / (next_y - vertex_y + 1e-9)
+    crosses = straddles & (x < slope + vertex_x)
+    return (jp.sum(crosses, axis=-1) % 2) == 1
+
+
+def fill_polygon(image, polygon, color):
```
oarriaga: We should look to move this inside a paz backend or in a new backend.

paz/backend/image.py
```diff
@@ -403,6 +399,86 @@ def random_color_transform(key, image):
...
+    """Fills a polygon in an image (JAX replacement for cv2.fillPoly)."""
+    grid_y, grid_x = jp.meshgrid(jp.arange(image.shape[0]),
+                                 jp.arange(image.shape[1]), indexing="ij")
+    inside = points_in_polygon(grid_x, grid_y, polygon)
+    color = paz.cast(color, image.dtype)
+    return jp.where(inside[..., jp.newaxis], color, image)
+
+
+def add_occlusion(key, image, num_vertices=6, max_radius_scale=0.5):
```
oarriaga: We should move this inside paz to a backend function.

paz/backend/image.py
```diff
@@ -403,6 +399,86 @@ def random_color_transform(key, image):
...
+    mask = jp.asarray(mask, jp.float32)
+    if mask.ndim == 2:
+        mask = mask[..., jp.newaxis]
+    foreground = paz.cast(image, jp.float32)
+    background = paz.cast(background, jp.float32)
+    blended = mask * foreground + (1.0 - mask) * background
+    return paz.cast(blended, jp.uint8)
+
+
+def points_in_polygon(grid_x, grid_y, polygon):
```
oarriaga: We should look to move this inside a backend file or a new backend file.

paz/backend/image.py
```diff
@@ -403,6 +399,86 @@ def random_color_transform(key, image):
...
+    return random_color_transform(keys[2 + num_occlusions], image)
+
+
+def sample_background(key, shape, backgrounds):
+    if backgrounds is None:
+        return make_random_plain_image(key, shape)
+    index = jax.random.randint(key, (), 0, len(backgrounds))
+    return backgrounds[index]
+
+
```
oarriaga, line 435: We should look into moving this affine transform into a backend function or a new backend file.

examples/implicit_orientation_learning/demo.py
oarriaga: Done in ed62c463 — both demo.py and train.py now use the classic `if __name__ == "__main__":` guard with the body inlined under it (no nested `main()`).

paz/backend/image.py
oarriaga: Searched all backends: the closest is `paz.draw.overlay_masks` (blends class-color masks, not a foreground-over-background composite), so no exact match. Moved `blend_background` into `paz.backend.draw` next to it.

paz/backend/image.py
oarriaga: Moved `points_in_polygon` into `paz.backend.draw` (with `fill_polygon`).

paz/backend/image.py
oarriaga: Moved `fill_polygon` into `paz.backend.draw` (drawing primitive, next to `overlay_masks`/`dot`). Test moved to draw_test.py.

paz/backend/image.py
oarriaga: Moved `add_occlusion` into `paz.backend.draw`; `randomize_rendered_image` now calls `paz.draw.add_occlusion`.

paz/backend/image.py
oarriaga: `affine_transform` is pre-existing (not added in this PR) and is imported by `paz.models.attention.stn` via `paz.image.affine_transform`. I left it in place to avoid mixing a cross-cutting relocation (with a caller update) into this domain-port PR — happy to move it to a geometry backend in a dedicated follow-up if you want.

paz/backend/draw.py
```diff
@@ -427,3 +428,49 @@ def text(image, message, point, scale=0.7, color=WHITE, thickness=2):
     font = cv2.FONT_HERSHEY_DUPLEX
     cv2.putText(image, message, point, font, scale, color, thickness)
     return image
+
+
+def blend_background(image, background, mask):
```
oarriaga, line 433: This might be better in the mask backend.

PR #366, opened 2026-06-24
Add paz.transformers toolkit and port Whisper decoding onto it

oarriaga:
## What

Introduces `paz/models/transformers` — surfaced as **`paz.transformers`** — a
small, functional toolkit of reusable sequence-model primitives, and ports the
Whisper example (`examples/speech_to_text`) onto it. First step toward sharing
transformer machinery across Whisper and Gemma4 (Gemma4 follows in a later PR).

### `paz.transformers` (new)

Module names the object/process; functions are bare verbs; params are explicit
(no `*Args` bundles). Each with a co-located `*_test.py`.

- `cache.build` / `cache.update` — KV cache `(batch, layers, 2, len, heads, dim)`
- `search.build` — jitted autoregressive decode loop; selectors `search.greedy`
  and `search.build_sampler(temperature, top_k, top_p)`
- `logits.apply_temperature` / `apply_top_k` / `apply_top_p`
- `mask.causal`
- `feedforward.gelu` (vanilla) / `feedforward.glu` (gated GeGLU)

### Whisper refactor (behavior-preserving)

`decoding.py` and `layers/{attention,decoder,encoder}.py` now use the toolkit.
QKV projection, cross-attention, learned positions and the warmup/prefill stay
model-side. Greedy decoding is unchanged — generated token ids are **identical**
before and after on `whisper_tiny_en` and `whisper_base_en` (on `harvard.wav`).
The demo gains optional `--temperature/--top_k/--top_p` sampling (greedy stays
the default) via `paz.transformers.search.build_sampler`.

### Incidental fix

`model.py` imported `ReversibleEmbedding` from `keras.layers`, which is not
present on the current keras; corrected to `keras_hub.layers`, matching
`examples/gemma4`, so the example loads.

## Validation

- Greedy token ids identical pre/post refactor (`whisper_tiny_en`,
  `whisper_base_en`).
- Sampling: `top_k=1` reduces to greedy; reproducible under a fixed key.
- `pytest paz/models/transformers/` → 13 passed.
- `pytest examples/speech_to_text/push_to_talk_test.py` → 10 passed.
  (Run with `KERAS_BACKEND=jax`, as for the rest of the JAX-first repo.)

## Notes

The branch is based on an earlier `paz-jax`; `paz/__init__.py` and
`examples/speech_to_text/` are unchanged on `paz-jax` since, so the diff is
conflict-free. Happy to rebase if preferred.

### DIFF
--- a/examples/speech_to_text/decoding.py
+++ b/examples/speech_to_text/decoding.py
@@ -8,10 +8,7 @@ from examples.speech_to_text.tokenizer import find_special_token_id
 PROMPT_TOKENS = ["<|startoftranscript|>", "<|transcribe|>", "<|notimestamps|>"]
 
 
-def kv_decode(decoder, cache_shape, cross_model, encoder_output, stop_id,
-              key=None):
-    if key is None:
-        key = jax.random.PRNGKey(0)
+def kv_decode(key, decoder, cache_shape, cross_model, encoder_output, stop_id):
     max_len = decoder.max_decode_length
     batch = int(encoder_output.shape[0])
     cache = build_self_cache(cache_shape, max_len, batch)
@@ -19,7 +16,7 @@ def kv_decode(decoder, cache_shape, cross_model, encoder_output, stop_id,
     cache = jp.asarray(cache)
     cross_cache = jp.asarray(cross_cache)
     stop = jp.array(stop_id, dtype=jp.int32)
-    buffer, length = decoder(cache, cross_cache, stop, key)
+    buffer, length = decoder(key, cache, cross_cache, stop)
     return ops.convert_to_numpy(buffer[0, :length]).tolist()
 
 
@@ -39,7 +36,7 @@ def KVDecoder(decoder_step, prompt_ids, max_tokens, max_seq=448, select=None):
     prompt_len = len(prompt_ids)
 
     @jax.jit
-    def decode(self_cache, cross_cache, stop_id, key):
+    def decode(key, self_cache, cross_cache, stop_id):
         buffer = jp.zeros((1, max_len), dtype=jp.int32)
         buffer = buffer.at[0, :prompt_len].set(prompt)
         cache = self_cache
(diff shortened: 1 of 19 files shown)


examples/speech_to_text/decoding.py
```diff
@@ -2,97 +2,75 @@
...
 
+import paz
 from examples.speech_to_text.tokenizer import find_special_token_id
 
 PROMPT_TOKENS = ["<|startoftranscript|>", "<|transcribe|>", "<|notimestamps|>"]
 
 
-def kv_decode(decoder, cache_shape, cross_model, encoder_output, stop_id):
+def kv_decode(decoder, cache_shape, cross_model, encoder_output, stop_id,
+              key=None):
```
oarriaga: We should always get a key.

examples/speech_to_text/demo.py
```diff
@@ -36,10 +39,10 @@
 MAX_AUDIO_SECONDS = 30
 
 
-def transcribe(wav_path, models, max_tokens=64):
+def transcribe(wav_path, models, max_tokens=64, select=None, key=None):
```
oarriaga: We should always pass a valid key not None. Also, key should be usually the first argument of a function.

examples/speech_to_text/decoding.py
oarriaga: Done in 40130157 — `kv_decode` now requires `key` as its first argument; dropped the `key=None` default and the `PRNGKey(0)` fallback. Same for the jitted `decode` and `search.build`'s `run`.

examples/speech_to_text/demo.py
oarriaga: Done in 40130157 — the demo always builds `key = jax.random.PRNGKey(args.seed)` (greedy ignores it), and `key` is now the first argument across the transcribe chain. No more `None` keys.

PR #370, opened 2026-06-24
Replace cv2.getAffineTransform with an explicit affine solve

oarriaga:
Cluster D (final) of the post-#365 cleanup. `human_pose_estimators.py` used cv2
for two things; this addresses the cleanly-replaceable one.

## Change
`build_affine` now computes the 3-point affine with a small numpy linear solve
(`solve_affine`) instead of `cv2.getAffineTransform`. This removes cv2 from the
keypoint back-projection path (`transform_to_image`, which only did
`getAffineTransform` + `np.dot`). The solved 2x3 matrix matches cv2 to ~1e-13,
so behavior is bit-exact.

## Why cv2.warpAffine stays
`cv2.warpAffine` warps the input into a **different output size** (resize+crop
into the model input) — `paz.image.affine_transform` only supports same-size
output, so it is not a drop-in. It is resize-class I/O (allowed by the guide)
and bit-exact critical for the HRNet pipeline that was verified against the
reference. Replacing it would require a new output-sized JAX warp and would
break that bit-exactness for no real gain (host preprocessing, once per frame).
Happy to do that as a separate, re-verified change if wanted.

## Verification (CPU)
- `solve_affine` vs `cv2.getAffineTransform`: max abs diff ~1e-13.
- `HigherHRNetHumanPose2D` on the dinner scene: 23 people, 17 keypoints each.
- lint + `pyflakes` clean.

This completes the flagged clusters (A inline-main, B JAX augmentation,
C EfficientDet postprocess, D pose affine).

### DIFF
--- a/paz/applications/human_pose_estimators.py
+++ b/paz/applications/human_pose_estimators.py
@@ -91,7 +91,14 @@ def build_affine(center, scale, size, inverse):
     source, destination = source_destination_points(center, scale, size)
     if inverse:
         source, destination = destination, source
-    return cv2.getAffineTransform(source, destination)
+    return solve_affine(source, destination)
+
+
+def solve_affine(source, destination):
+    source = np.asarray(source, np.float64)
+    destination = np.asarray(destination, np.float64)
+    homogeneous = np.concatenate([source, np.ones((3, 1))], axis=1)
+    return np.linalg.solve(homogeneous, destination).T
 
 
 def source_destination_points(center, scale, size, factor=200):


paz/applications/human_pose_estimators.py
```diff
@@ -91,7 +91,14 @@ def build_affine(center, scale, size, inverse):
     source, destination = source_destination_points(center, scale, size)
     if inverse:
         source, destination = destination, source
-    return cv2.getAffineTransform(source, destination)
```
oarriaga, line 94: Let's keep warp affine for now

oarriaga:
Closing per review ('keep warp affine for now'). The only change here was swapping cv2.getAffineTransform for a numpy solve, which still feeds cv2.warpAffine — a lateral cv2->numpy move, not the JAX goal, on bit-exact-verified HRNet code. Folding the whole affine (getAffineTransform + warpAffine) into a single, properly-verified JAX port later instead.

PR #371, opened 2026-06-24
Adopt paz.transformers in Gemma4: rotary, RMSNorm, logit transforms

oarriaga:
## What

Second step of the `paz.transformers` rollout (after #366, Whisper): point
**Gemma4** at the shared library and move its remaining reusable pieces in.
Behavior-preserving — every gemma4 decode/model/sampling/inference test stays
identical.

### Moved into the library

- **`RMSNormalization`, `MergeDims`, `SplitDim` → `paz.layers`** (de-prefixed
  from `Gemma4RMSNormalization`; ported verbatim). Gemma4's attention GQA
  reshapes and the RoPE rotation now use the shared layers.
- **Rotary embedding → `paz.transformers.embeddings.rotary`** —
  `rotary.apply` / `apply_partial` / `build` / `build_frequencies`, ported from
  `examples/gemma4/layers/core.py`. (`embeddings/` is a subpackage so other
  positional encodings can join later.)
- **`logits` reuse** — gemma4 `sampling.py` now composes
  `paz.transformers.logits.apply_temperature/top_k/top_p` instead of its own
  copies (identical code).

`examples/gemma4/layers/core.py` loses ~138 lines (rotary + the two reshape
layers) now that they live in the library. Co-located unit tests added for the
new library pieces.

## Validation

- gemma4 oracle unchanged: `pytest examples/gemma4/{sampling,decoding,multimodal_decoding,inference,model,causal_lm}_test.py` → 26 passed (same as before this PR).
- New unit tests: `pytest paz/models/transformers/ paz/layers/` → green
  (rotary identity-at-position-0, RMSNorm vs reference, merge/split reshapes).
- Whisper (#366) still imports the changed `paz.layers` / `paz.transformers`
  `__init__`s cleanly: `push_to_talk_test` green.
- All runs use `KERAS_BACKEND=jax`. `pyflakes` clean (only the usual
  `__init__` re-export notices).

## Deferred to a follow-up

These are the genuinely divergent / higher-risk bits, kept out to keep this PR
behavior-preserving and reviewable:

- **KV-cache `update` reconciliation** — Gemma splits K/V (`cache[:, 0]` /
  `cache[:, 1]`, update at `[0, index, 0, 0]`) while the shared `cache.update`
  keeps them stacked (`[0, 0, index, 0, 0]`); unifying needs a careful
  bit-for-bit pass.
- **Stateless decode helpers** (`snapshot_variables`, `call_stateless`) →
  `paz.backend.standard`.
- **`feedforward.glu`** adoption in gemma4's gated FFN, **`reshape_query`**, and
  unifying gemma4's jitted decode loop with `paz.transformers.search`
  (gemma4 threads weights as jit args; Whisper closes over the model).


examples/gemma4/layers/attention.py
```diff
@@ -63,7 +64,7 @@ def cached_attend(args, shared_kv_cache=None):
         args.partial_rotary,
         positions,
     )
```
oarriaga: This is really hacky. We should not use named tuples for hidding arguments. If a function requires so many arguments we should split it and if it's a very high level function that does require so many inputs then one should embrace that complexity instead of hiding it.

oarriaga: Fixed in d1265b08 — removed `AttendArgs`/`CachedAttendArgs`. `attend`/`cached_attend`/`compute_attention` now have explicit signatures listing every input; callers build a plain tuple and unpack with `fn(*args)` (the CLAUDE.md many-args pattern), so nothing is hidden behind `args.field`. Updated decoder.py and vision.py, and dropped a stale unused import. gemma4 decode/model/vision/sampling tests stay identical.

PR #376, opened 2026-06-25
Add Cityscapes 8-category loader, training, and demo

oarriaga:
Closes the remaining semantic-segmentation gap from the roadmap: a real
on-disk **Cityscapes** path on top of the existing UNet example.

## What this adds
- **`paz/datasets/cityscapes.py`** — loader returning paired
  `leftImg8bit`/`gtFine` paths (lazy, mirroring `voc.load`) plus per-sample
  `load_image`/`load_mask` helpers. `build_id_to_category()` is a length-34 LUT
  built from `CATEGORY_RANGES` that remaps the raw label IDs to the **8 official
  Cityscapes `categoryId` groups**: `void, flat, construction, object, nature,
  sky, human, vehicle`. This is faithful to legacy master's
  `CITY_ESCAPES_ID_TO_MASK`. `void` is class 0 (a real trained channel), so
  `paz.losses.dice` applies directly with **no ignore-index machinery**.
- **`paz/datasets/__init__.py`** — append-only registration in `load()` and
  `labels()` (`name="Cityscapes"`).
- **`examples/semantic_segmentation/train_cityscapes.py`** — `UNET_VGG16`
  (softmax) trained via a `keras.utils.PyDataset` that loads per batch, resizing
  images bilinear and labels nearest-neighbor and one-hotting the 8 categories.
  Uses Cityscapes' own `val` split for validation.
- **`examples/semantic_segmentation/demo_cityscapes.py`** — predict → argmax →
  `paz.draw.overlay_masks` colored overlay.
- **`cityscapes_test.py`**, README section, and roadmap update.

## Why 8 categories
The legacy master scheme maps the 34 raw IDs to exactly the official
`categoryId` grouping (verified ID-by-ID), so this is principled, not a
simplification hack. It also sidesteps the ignore-index problem entirely.

## Verification (CPU, `KERAS_BACKEND=jax`)
- `pytest paz/datasets/cityscapes_test.py` — 3 passed (ID mapping, path
  pairing, image/mask shapes & value range).
- `pytest paz/losses/segmentation_test.py paz/models/segmentation/unet_test.py`
  — 6 passed (no regressions).
- **Train smoke** on a synthetic fake-Cityscapes tree (`--backbone_weights ""`,
  `--image_size 128`, 3 epochs): dice loss decreases monotonically
  (0.889 → 0.880 → 0.874), weights written.
- **Demo smoke**: loads those weights, runs on a synthetic `leftImg8bit` PNG,
  writes a valid colored overlay.

Cityscapes data is login-gated, so no weights are hosted — train to produce
them. Real-data verification was done on a synthetic tree matching the official
directory layout.

### DIFF
--- a/examples/semantic_segmentation/train_cityscapes.py
+++ b/examples/semantic_segmentation/train_cityscapes.py
@@ -45,8 +45,11 @@ if __name__ == "__main__":
     parser.add_argument("--learning_rate", default=0.001, type=float)
     parser.add_argument("--epochs", default=100, type=int)
     parser.add_argument("--backbone_weights", default="imagenet")
+    parser.add_argument("--download", action="store_true")
     args = parser.parse_args()
     os.makedirs(args.save_path, exist_ok=True)
+    if args.download:
+        cityscapes.download(args.root)
     H = W = args.image_size
     num_classes = len(cityscapes.get_class_names())
     train_paths = cityscapes.load(args.root, "train")
--- a/examples/semantic_segmentation/train_cityscapes.py
+++ b/examples/semantic_segmentation/train_cityscapes.py
@@ -1,0 +1,72 @@
+import os
+
+os.environ["KERAS_BACKEND"] = "jax"
+
+import argparse
+
+import numpy as np
+import keras
+
+import paz
+from paz.datasets import cityscapes
+
+
+class CityscapesSequence(keras.utils.PyDataset):
+    def __init__(self, image_paths, label_paths, num_classes, size, batch_size):
+        super().__init__()
+        self.image_paths = image_paths
+        self.label_paths = label_paths
+        self.num_classes = num_classes
+        self.size = size
+        self.batch_size = batch_size
+        self.mean = np.array(paz.image.BGR_IMAGENET_MEAN, "float32")
+
+    def __len__(self):
+        return len(self.image_paths) // self.batch_size
+
+    def __getitem__(self, index):
+        chunk = slice(index * self.batch_size, (index + 1) * self.batch_size)
+        images, labels = [], []
+        for image_path in self.image_paths[chunk]:
+            images.append(cityscapes.load_image(image_path, self.size))
+        for label_path in self.label_paths[chunk]:
+            labels.append(cityscapes.load_mask(label_path, self.size))
+        images = np.asarray(images, "float32")[..., ::-1] - self.mean
+        targets = np.eye(self.num_classes, dtype="float32")[np.asarray(labels)]
+        return images, targets
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description="UNet Cityscapes segmentation")
+    parser.add_argument("--root", required=True, help="Cityscapes directory")
+    parser.add_argument("--save_path", default="experiments")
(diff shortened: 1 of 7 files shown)


paz/datasets/cityscapes.py
```diff
@@ -0,0 +1,57 @@
+import os
+from glob import glob
+
+import numpy as np
+
+import paz
+
+
```
oarriaga, line 13: We should also try to make a download function if available.

examples/semantic_segmentation/train_cityscapes.py
```diff
@@ -0,0 +1,72 @@
+import os
+
+os.environ["KERAS_BACKEND"] = "jax"
+
```
oarriaga, line 4: We should try to download and test and get positive results including a good predictive mask labels as in the paz main branch example image.

paz/datasets/cityscapes.py
oarriaga: Added `cityscapes.download(root)` in 9d01d9ea. It uses the maintained `cityscapesscripts` downloader (lazy optional import + clear error, mirroring how `kaggle_utils` handles the `kaggle` dep) and reuses `kaggle_utils.extract_archive`. It reads CITYSCAPES_USERNAME / CITYSCAPES_PASSWORD from env so it works headless. Cityscapes has no anonymous URL, so credentials from a free account are unavoidable.

examples/semantic_segmentation/train_cityscapes.py
oarriaga: `train_cityscapes.py` now takes `--download`, which calls `cityscapes.download(args.root)` before training. I verified the full pipeline end-to-end on a synthetic fake-Cityscapes tree (loader pairing + ID→category remap tests pass; 3-epoch offline train shows dice loss decreasing; demo writes a colored overlay). I can't run the real download + train-to-good-masks here because Cityscapes is login-gated and I have no credentials installed — see my note in the PR thread on how to let me complete that.

oarriaga: Done — trained on **real Cityscapes** and got good masks. Data fetched via a Kaggle mirror that carries the official `_gtFine_labelIds.png`; added `download_kaggle(root)` (+ `adapt_kaggle_layout`) in 5f75af06 alongside the official `download()`.

**UNET_VGG16** (ImageNet backbone, softmax, Dice loss), 35 epochs @ 256px on an RTX A4000, evaluated on the full 500-image val split:

| metric | value |
|---|---|
| val pixel accuracy | 89.3% |
| **val mean IoU (8 categories)** | **0.689** |
| flat / sky / nature | 0.91 / 0.84 / 0.82 |
| vehicle / construction | 0.78 / 0.75 |
| void / human / object | 0.60 / 0.50 / 0.32 |

(human/object lag because they're thin/small at 256px — expected.) The committed loader ran unchanged on the real data: exact official 2975/500 train/val counts, all 8 categories present, label-path pairing correct.

@oarriaga I'll attach the predicted-mask composite (input · ground-truth · prediction, 4 val scenes) to the PR shortly — it's saved locally. Trained weights (285 MB) are available if you'd like them hosted on altamira-data for a demo checkpoint.

PR #377, opened 2026-06-25
Add Whisper speech-to-text foundation model

oarriaga:
## What

Second PR of promoting Whisper into paz. Adds the Whisper speech-to-text
**foundation model** under `paz/models/foundation/whisper/`, following the
dinov3 layout (`layers/` + `models/`).

- `models/whisper.py`: `WhisperFrontend`, `WhisperEncoder`, `WhisperCrossCache`,
  `WhisperDecoderStep` (functional `keras.Model`s with trainable layers), plus a
  `Whisper(model_name, weights="paz")` bundle returning the four models.
  Surfaced as `paz.models.Whisper`.
- `layers/`: attention, encoder, decoder, embedding, frontend, utils — moved
  from the example, reusing `paz.transformers.{cache,search,mask,feedforward}`.
- `configuration.py` (`CONFIGS` + `to_model_args`), `decoding.py` (KV decode
  loop over `paz.transformers.search`), `tokenizer.py` (byte-level decoder).
- Pretrained weights load by name via `keras.utils.get_file` (the detector
  standard, e.g. `haar_cascade.py`), with a `models_path` override for local
  use until the weights are uploaded. Tokenizer assets resolve the same way.

## Progressive disclosure

This is the middle layer: a user can build the models, fine-tune them (standard
trainable Keras graphs), or recombine the layers/configs. The next PR adds the
one-call `paz.applications.TranscribeWhisper`.

## Validation

- `pytest paz/models/foundation/whisper/whisper_test.py` -> 4 passed.
  Includes a **bit-for-bit token-id snapshot** on `harvard.wav` matching the
  pre-move example output exactly (the tokenizer-free oracle), plus build/shape
  and special-token contracts. Weight/asset-dependent tests skip on CI.
- Regression: `pytest paz/models/transformers/ paz/layers/ paz/backend/audio_test.py`
  -> 29 passed. `KERAS_BACKEND=jax`, lines <= 80, pyflakes clean.

## Note on tokenizer assets

Text decoding needs the authoritative flat `vocabulary.json` (id<->token map).
The keras_hub-style assets currently available are `merges.txt` + the special-
token config, but `vocabulary.json` (the decode-critical sibling) is still
missing, and it cannot be reconstructed exactly from this `merges.txt`. The
token-id pipeline (and this PR's oracle) is fully validated; text output will
light up once `vocabulary.json` is provided/uploaded alongside the weights.

## Roadmap (3 PRs)

- #375 `paz.audio` backend — merged
- **this PR** — Whisper foundation model (`paz.models.Whisper`)
- next — `paz.applications.TranscribeWhisper` + rewire the example onto paz


paz/models/__init__.py
```diff
@@ -32,4 +32,5 @@
 from .autoencoder.autoencoder import extract_encoder
 from .foundation.dinov3.models.vision_transformer import DINOV3VITS
 from .foundation.dinov3.models.vision_transformer import DINOV3VITB
-from .foundation.dinov3.models.vision_transformer import DINOV3VITL
\ No newline at end of file
+from .foundation.dinov3.models.vision_transformer import DINOV3VITL
+from .foundation.whisper.models.whisper import Whisper
```
oarriaga: These are too many files. foundation/whisper.models.whisper the structure should be more flat.

paz/models/transformers/attention.py
```diff
@@ -0,0 +1,171 @@
+import string
+
+import paz
+from keras import ops
+from keras.layers import Dropout, EinsumDense
+
+from paz.models.foundation.whisper.layers.utils import Kernel
+
+
+def attend(query, value, num_heads, key_dim, dropout, name):
```
oarriaga, line 10: I see that a lot of functions are not using the paz.transformers. modules we have already ported. For example there are already attention layers / functions. We should re use that as much as possible for all functions here in this file.

paz/models/foundation/whisper/layers/decoder.py
```diff
@@ -0,0 +1,53 @@
+import paz
+from keras import ops
+from keras.layers import LayerNormalization, Dropout
+
+from paz.models.foundation.whisper.layers.attention import kv_attend
+
```
oarriaga, line 6: I see that a lot of functions are not using the paz.transformers. modules we have already ported. For example there are already attention layers / functions. We should re use that as much as possible for all functions here in this file.

paz/models/transformers/embeddings/absolute.py
```diff
@@ -0,0 +1,46 @@
+from keras import ops
+from keras.layers import Embedding, Lambda
+
+from paz.models.foundation.whisper.layers.utils import Kernel
+
```
oarriaga, line 5: We should try to reuse as much of the paz.transformers module and also abstract whatever possible there in case some of these functions might be useful for later models and not only to whisper.

paz/models/foundation/whisper/layers/frontend.py
```diff
@@ -0,0 +1,121 @@
+import numpy as np
+from keras import ops
+
```
oarriaga: The frontend might be useful for Gemma4 as well thus we should also abstract this in a module above that can be reused across many layers / models.

paz/models/foundation/whisper/layers/utils.py
```diff
@@ -0,0 +1,21 @@
+import keras
+from keras.activations import gelu
+from keras.layers import Dense
+
```
oarriaga: We are not using the paz.transformers module

paz/models/foundation/whisper/model.py
```diff
@@ -0,0 +1,249 @@
...
+from paz.models.foundation.whisper.layers.encoder import encoder_block
+from paz.models.foundation.whisper.layers.decoder import decoder_block
+from paz.models.foundation.whisper.configuration import to_model_args
+
+WHISPER_WEIGHTS_URL = "https://github.com/oarriaga/altamira-data/releases/download/v0.20/whisper/"  # fmt: skip
+DECODER_LAYER = "transformer_decoder_layer_{}"
+NORM_KWARGS = {"axis": -1, "epsilon": 1e-5}
+
+
+def Whisper(model_name="whisper_base_en", weights="paz", models_path=None):
```
oarriaga, line 26: My impression is that Whisper should return a single structure either a full Keras model that can be called end-to-end or a namedtuple that contains all these elements. But that depends on how the model us used later on. This has to be discussed in detail.

paz/models/foundation/whisper/model.py
```diff
@@ -0,0 +1,249 @@
+from pathlib import Path
+
```
oarriaga, line 3: This file is too deep into different modules. We should have a more flat hierarchy that does not make foundation.whisper.models.whisper. but maybe paz.foundation.whisper.models

paz/models/foundation/whisper/decoding.py
```diff
@@ -0,0 +1,88 @@
+import jax
+import jax.numpy as jp
+from keras import ops
+
+import paz
+from paz.models.foundation.whisper.tokenizer import find_special_token_id
+
+PROMPT_TOKENS = ["<|startoftranscript|>", "<|transcribe|>", "<|notimestamps|>"]
+
+
```
oarriaga, line 10: Same as before we should re-use all our compoments from paz.transformers or integrate it by abstracting what could be reusable.

paz/models/foundation/whisper/tokenizer.py
```diff
@@ -0,0 +1,113 @@
+import json
+from pathlib import Path
+
+from keras.utils import get_file
+
```
oarriaga, line 5: If this tokenizer can be reused for other models we should abstract it in a higher level module paz.transformers.tokenizers (or similar) if not we can leave it here.

oarriaga:
Thanks for the detailed review — addressed in the latest commit. Summary of how each comment was handled:

**Reuse / abstract into paz.transformers** (attention.py, decoder.py, embedding.py, utils.py)
- New `paz.transformers.attention`: generic EinsumDense multi-head attention (`attend`, `kv_attend`, `build_cache`/`update_cache`, projections), extracted from the Whisper layer. `kv_attend` now takes explicit params (no `config` bundle).
- New `paz.transformers.embeddings.absolute`: learned positional embeddings (sibling of `rotary`).
- `utils.py` removed: the shared `TruncatedNormal(0.02)` initializer is now `paz.transformers.attention.kernel`; GELU dense already comes from `paz.transformers.feedforward`.
- Whisper's `layers/` now holds only the model-specific encoder/decoder **block assembly**, which consumes the shared attention + `mask` + `feedforward` + `cache` + `search`.

**Frontend reusable across models** (frontend.py)
- Mel log-spectrogram (STFT + mel filterbank) moved into `paz.audio` (`log_mel_spectrogram`, `build_mel_filters`) so audio models like Gemma can reuse it.

**Tokenizer abstraction** (tokenizer.py)
- Byte-level BPE decode core (GPT-2 family) lifted to `paz.transformers.tokenizers`. Whisper keeps only its special-token map + asset resolution.

**Flatter hierarchy** (models/__init__.py, model.py)
- Dropped `models/whisper.py`; builders now live in `paz/models/foundation/whisper/model.py`. Surfaced as `paz.models.Whisper`.

**Return structure** (model.py)
- `Whisper(...)` now returns a namedtuple `Whisper(frontend, encoder, cross_cache, decoder_step)` (tuple-unpacking and indexing still work). The autoregressive KV-cache loop is a `jax.lax.while_loop`, so a single end-to-end static graph doesn't fit; the namedtuple keeps each part trainable/swappable.

**Decoding** — already built on `paz.transformers.{cache, search, mask}`; now also benefits from the shared attention. The decode loop is unchanged.

**gemma4** — left as-is for now per discussion: its attention (RoPE + GQA + soft-cap) differs enough that a single shared function would need behavior-switching flags. It can adopt the shared core later when bit-for-bit safe.

**Validation** — the `harvard.wav` token-id snapshot still matches **bit-for-bit** (weights load by identical layer names), confirming the refactor is behavior-preserving. New co-located tests for `attention`, `tokenizers`, `embeddings.absolute`, and the audio frontend. `pytest paz/models/transformers/ paz/backend/audio_test.py paz/models/foundation/whisper/whisper_test.py` → 34 passed.

Still pending (separate from this refactor): the authoritative flat `vocabulary.json` is needed for id→text decoding; it can't be reconstructed exactly from the lossy `merges.txt`. The token-id pipeline is fully validated meanwhile.

paz/models/__init__.py
oarriaga: Flattened. Dropped `models/whisper.py`; the builders now live in `paz/models/foundation/whisper/model.py` (still surfaced as `paz.models.Whisper`). Most of the files were generic layers — those moved into `paz.transformers`, so the whisper module now holds only `model.py`, `layers/` (encoder/decoder block assembly), `decoding.py`, `tokenizer.py` and `configuration.py`.

paz/models/transformers/attention.py
oarriaga: This file is now `paz.transformers.attention` — the generic multi-head attention (attend, kv_attend, build/update_cache, projections) extracted out of Whisper so it is reusable. It uses `paz.transformers.cache` for the KV-cache writes, and Whisper's encoder/decoder now import `attend`/`kv_attend` from here instead of reimplementing them. `kv_attend` also takes explicit params now (no config bundle).

paz/models/foundation/whisper/layers/decoder.py
oarriaga: Done. `decoder.py` now imports `kv_attend` from `paz.transformers.attention` and keeps only Whisper's block assembly — self + cross attention + FFN, with the causal mask via `paz.transformers.mask.causal` and the FFN via `paz.transformers.feedforward.gelu`.

paz/models/transformers/embeddings/absolute.py
oarriaga: Abstracted. This is now `paz.transformers.embeddings.absolute` (sibling of `rotary`) — the learned positional embedding lifted out of Whisper so future models can reuse it. It reuses the shared `paz.transformers.attention.kernel` initializer.

paz/models/foundation/whisper/layers/frontend.py
oarriaga: Agreed — moved to `paz.audio` as `log_mel_spectrogram` + `build_mel_filters` (in `paz/backend/audio.py`), so audio models such as Gemma can reuse the frontend. The old `layers/frontend.py` is removed.

paz/models/foundation/whisper/layers/utils.py
oarriaga: Removed `utils.py` entirely. The shared `TruncatedNormal(0.02)` initializer is now `paz.transformers.attention.kernel`, and the GELU dense comes from `paz.transformers.feedforward`.

paz/models/foundation/whisper/model.py
oarriaga: `Whisper(...)` now returns a namedtuple `Whisper(frontend, encoder, cross_cache, decoder_step)` (tuple-unpacking and indexing still work). A single end-to-end Keras model does not fit cleanly because generation is a `jax.lax.while_loop` over the KV cache, not a static graph; the namedtuple keeps each part trainable/swappable for fine-tuning. Happy to revisit if you want a `.generate()` wrapper later.

paz/models/foundation/whisper/model.py
oarriaga: Flattened to `paz/models/foundation/whisper/model.py` (removed the `models/whisper.` double-name). I kept it under `paz.models.foundation` for parity with `dinov2`/`dinov3`. If you would rather promote the whole foundation tree to a top-level `paz.foundation`, I can do that as a separate change across all three models.

paz/models/foundation/whisper/decoding.py
oarriaga: `decoding.py` already drives `paz.transformers.search` and builds the KV cache via `paz.transformers.cache`; with this change the per-step model also runs on the shared `paz.transformers.attention`. What remains here is Whisper-specific orchestration: the prompt warm-up `fori_loop` and the cross-attention cache wiring.

paz/models/foundation/whisper/tokenizer.py
oarriaga: Abstracted. The reusable part — byte-level BPE decoding (GPT-2 family: char/byte maps, vocabulary load, id->bytes decode) — is now `paz.transformers.tokenizers`. `tokenizer.py` keeps only the Whisper-specific bits: the special-token map and asset (vocabulary/merges/config) resolution.

PR #385, opened 2026-07-01
Fast streaming generation for foundation models, with API and module cleanup

oarriaga:
## Summary

Adds a shared streaming decode primitive and a compile-once, padded generator
for Gemma4 (text and image), then cleans up the foundation-model APIs per
review. Four commits, each a self-contained step.

### 1. Fast streaming generation
- `paz.models.transformers.search`: `build_streaming(step, select, max_tokens,
  max_length, emit)` replaces `build`; `emit(token_id)` fires per token via an
  ordered `jax.debug.callback` so any decoder can stream. `discard` is the
  silent sink, `emit_token` the shared helper.
- Whisper and Gemma4 KVDecoders stream through it; `TranscribeWhisper` streams
  live via an incremental UTF-8 printer (safe across multi-byte splits).
- Gemma4 `build_generator(prompt_ids, vision_embeddings, vision_indices)`:
  right-pads prompts over a fixed cache so prefill and the decode loop compile
  once and are reused; handles text and image uniformly. `build_prompt_rows`
  vectorized to one batched embedding lookup. `GenerateGemma4`/
  `DescribeImageGemma4` stream; the image factory exposes `encode`/`answer` so
  an image is encoded once and reused across questions.

### 2. API naming, weights, device (review feedback, applied generally)
- Drop the redundant package name from module-internal symbols
  (`resolve_gemma4_dir` -> `resolve_dir`, `decode_whisper_tokens` ->
  `decode_tokens`, etc.); flat public names keep their model identity.
- Rename the pretrained-weights selector `"paz"` -> `"pretrained"` across
  gemma4, whisper, dinov3 (these are the official pretrained weights paz only
  hosts).
- Add `model_device` + a pytree-general `place_on_model_device` to
  `paz.backend.standard` (exported from `paz`); `encode_image` is now
  device-neutral and applications accept a pre-built `models` bundle, so the
  caller chooses CPU vs GPU at construction. Add `DescribeImageGemma42B/44B`.

### 3. Fold `inference.py` + `pretrained.py` into `model.py`
- Matches the Whisper template: the whole Gemma4 model (bundle factory, step
  models, weight loading) now lives in `model.py`, imported as
  `from paz.models.foundation.gemma4.model import Gemma4`. Removes the odd
  `pretrained` module and the cross-module import layering.

## Measured (RTX A4000, gemma4_2b)
- Gemma4 image decode: 44.4 s -> 2.3 s (~19x), bit-for-bit identical tokens.
- Vision encoder stays fp32 by design (matches keras-hub / the reference) and
  runs on CPU on a 16 GB GPU; the demo shows the build-on-CPU recipe.

## Testing
`KERAS_BACKEND=jax JAX_PLATFORMS=cpu pytest` on gemma4 + transformers + whisper:
80 passed, 4 skipped (weights-gated). pyflakes clean. New tests cover
bit-for-bit parity of the padded generator vs the eager reference and in-order
streaming.


examples/gemma4/demo_image.py
```diff
@@ -14,8 +14,13 @@
 if str(ROOT) not in sys.path:
     sys.path.insert(0, str(ROOT))
 
+import jax
+
 import paz
 from paz.applications import DescribeImageGemma4
+from paz.models import Gemma4
+from paz.models.foundation.gemma4.pretrained import (
+    resolve_gemma4_dir, load_gemma4_vision_encoder)
```
oarriaga: No need to name gemma4 since it's clear from the path paz.models.foundation.gemma4.pretrained that we are already taking about gemma4. Moreover I don't know if the module "pretrained" is really the right separation, maybe that should be all under the module gemma4?

paz/applications/generators.py
```diff
@@ -1,50 +1,79 @@
...
+from paz.models.foundation.gemma4.multimodal_decoding import (
+    build_generator, build_text_generator)
 from paz.models.foundation.gemma4.vision import VisionEncoderArgs
 
 
-def GenerateGemma4(model_name="gemma4_2b", max_tokens=64, weights="paz",
-                   models_path=None):
+def GenerateGemma4(model_name="gemma4_2b", max_tokens=64, max_seq=512,
+                   max_prompt=128, weights="paz", models_path=None,
+                   models=None):
```
oarriaga, line 19: These variables and their values are not great. GenerateGemma4 should have a factory that makes the model_name not be there anymore i.e. GenerateGemma42B. weights should not be paz since these are not "our" weights. I am not sure why we need models_path or models. If this can be used as a lower level function to build other Gemma4 models those variables might make sense, but not directly for a user interface.

paz/applications/generators.py
```diff
@@ -1,50 +1,79 @@
 import json
 from pathlib import Path
 
+import jax
 import numpy as np
 from keras import ops
 
 from paz.models import Gemma4
 from paz.models.foundation.gemma4.pretrained import resolve_gemma4_dir
```
oarriaga: Same as in the other file no need to call it "resolve_gemma4_dir" since gemma4 is already clear from the module path.

oarriaga: Apply this also everywhere where there is a similar case.

paz/applications/generators.py
```diff
@@ -1,50 +1,79 @@
...
+def build_token_printer(tokenizer, stop_id):
+    def print_token(token_id):
+        if int(token_id) != stop_id:
+            print(tokenizer.detokenize([int(token_id)]), end="", flush=True)
+    return print_token
+
+
+def DescribeImageGemma4(model_name="gemma4_2b", max_tokens=64, max_seq=512,
+                        max_prompt=400, weights="paz", models_path=None,
+                        models=None):
```
oarriaga, line 48: Same as the GenerateGemma4 function, the variables and values of some of these elements are strange i.e. model_name, models_path, models and "paz".

paz/applications/generators.py
```diff
@@ -62,6 +91,9 @@ def encode_image(image, models, vision):
     if image.max() > 1.0:
         image = image / 255.0
     inputs = preprocess_images(image[None], vision)
+    # Run the vision encoder wherever its weights live: placing the inputs on
+    # that device lets the caller decide CPU vs GPU by how they built the model.
+    inputs = place_on_model_device(inputs, models.vision_encoder)
```
oarriaga, line 96: This seems to be only applicable for the vision_encoder; however, a user might do the same for any element of the model.

paz/applications/transcribers.py
```diff
@@ -8,16 +10,20 @@
     KVDecoder, kv_decode, build_whisper_prompt_token_ids,
     extract_text_token_ids)
 from paz.models.foundation.whisper.tokenizer import (
-    find_special_token_id, decode_whisper_tokens)
+    find_special_token_id, decode_whisper_tokens, build_whisper_byte_maps)
```
oarriaga: Same comment as in the other module paths "whisper" might not be necessary given the module path.

PR #388, opened 2026-07-03
Fix SSD detector mAP eval under the torch backend

oarriaga:
## Summary

The SSD detector (`paz.applications.detectors.SSD`) fed the raw model output
straight into its JAX postprocess (`jp.squeeze`, `paz.detection.decode`, ...).
Under `KERAS_BACKEND=torch` the model emits **torch tensors**, so the JAX ops
rejected them:

```
TypeError: squeeze requires ndarray or scalar arguments, got <class 'torch.Tensor'>
```

This crashed the `EvaluateMAP` training callback during torch-backend training
(the callback had only ever been exercised standalone under the jax backend).

## Fix

Convert the model output with `keras.ops.convert_to_numpy` at the inference
boundary, so the postprocess consumes it on **either** backend (no-op-ish under
jax, torch→cpu→numpy under torch). One-line, backend-agnostic.

## Verified

- Detector + `compute_mAP` now run under `KERAS_BACKEND=torch` (previously
  crashed); unchanged under `KERAS_BACKEND=jax`.
- End-to-end: a training run with the mAP callback now completes the eval pass
  cleanly instead of crashing at the first `on_epoch_end`.

Base: `paz-jax`.

### DIFF
--- a/paz/applications/detectors.py
+++ b/paz/applications/detectors.py
@@ -3,6 +3,7 @@ import cv2
 import jax.numpy as jp
 import numpy as np
 import jax
+import keras
 from keras import ops
 import paz
 
@@ -34,9 +35,12 @@ def SSD(model, score_thresh, prior_boxes, variances, apply_NMS, draw):
 
     def call(image):
         image_size = paz.image.get_size(image)
-        # convert_to_numpy makes the model output backend-agnostic: the JAX
-        # postprocess consumes it whether the backend is jax or torch.
-        outputs = ops.convert_to_numpy(model(preprocess(image)))
+        outputs = model(preprocess(image))
+        # On the torch backend the model emits torch tensors, which the JAX
+        # postprocess cannot consume, so convert them. On the jax backend the
+        # output is already a jax array and flows through on-device.
+        if keras.backend.backend() == "torch":
+            outputs = ops.convert_to_numpy(outputs)
         detections = postprocess(outputs, image_size)
         detections = paz.detection.remove_invalid(detections)
         return paz.detection.to_boxes2D(detections)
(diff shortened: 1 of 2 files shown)


paz/applications/detectors.py
```diff
@@ -31,7 +34,10 @@ def postprocess(detections, image_size):
 
     def call(image):
         image_size = paz.image.get_size(image)
-        detections = postprocess(model(preprocess(image)), image_size)
+        # convert_to_numpy makes the model output backend-agnostic: the JAX
+        # postprocess consumes it whether the backend is jax or torch.
+        outputs = ops.convert_to_numpy(model(preprocess(image)))
```
oarriaga: This most likely adds an expensive overhead transferring data from GPU to CPU to GPU.

paz/callbacks/evaluate_map.py
```diff
@@ -28,6 +30,7 @@ def on_epoch_end(self, epoch, logs=None):
                   "iou_thresh": self.iou_thresh,
                   "use_07_metric": self.use_07_metric, "verbose": True}
         result = paz.evaluation.compute_mAP(*args, **kwargs)
+        gc.collect()  # reclaim the per-eval detection arrays on low-RAM hosts
```
oarriaga: Is this necessary after the jax.jit fix?

paz/applications/detectors.py
oarriaga: Fixed in cbd80c34: the conversion is now conditional on the torch backend. On the jax backend the model output is already a jax array, so it flows straight into the jax postprocess on-device -- no GPU->host->GPU round trip. Verified the mAP path runs cleanly on both backends.

paz/callbacks/evaluate_map.py
oarriaga: Removed in cbd80c34. The reclaim test showed gc.collect() freed ~0 (after eval 4.86GB -> after gc 4.86GB); the un-jit preprocess change is the real leak fix, so gc was dead weight.

PR #394, opened 2026-07-10
Drop keras_hub runtime dependency with an internal ReversibleEmbedding

oarriaga:
## What

`ReversibleEmbedding` from `keras_hub` was the only runtime use of keras_hub in
paz — imported by the Whisper and Gemma4 foundation models. This adds an
internal equivalent in `paz.models.transformers.embeddings.reversible` and
points both models at it.

After this, importing and running Whisper/Gemma4 no longer requires keras_hub.
It stays an optional dependency only for the offline weight-conversion tool
(`gemma4/conversion.py`) and the opt-in parity tests, which already
`pytest.importorskip` it and skip cleanly when it is absent.

## Scope

Only the ReversibleEmbedding dependency. TFP and Keras 3 are untouched.

The internal layer reproduces the non-quantized behavior both models actually
use: forward embedding lookup and a reverse projection to the vocabulary, tied
(default) or untied. The quantization / soft-cap / reverse-dtype machinery from
keras_hub is dropped as unused — Gemma applies its soft-cap externally via the
existing `logits.soft_cap`.

## Verification

- Bit-for-bit parity against `keras_hub.layers.ReversibleEmbedding`: forward
  exact, reverse max abs diff `0.0`, for both tied and untied weights.
- New co-located `reversible_test.py` (self-contained, no keras_hub).
- `pytest paz/models/transformers/embeddings/ whisper_test.py
  gemma4/model_test.py`: 16 passed, 3 skipped (weights-only), on
  `KERAS_BACKEND=jax`. Models build, save, and load with the internal layer.
- pyflakes clean.

### DIFF
--- a/paz/models/foundation/gemma4/model.py
+++ b/paz/models/foundation/gemma4/model.py
@@ -8,7 +8,7 @@ from keras import Model, ops
 from keras.initializers import VarianceScaling
 from keras.layers import Embedding, EinsumDense, Input
 from keras.utils import get_file
-from keras_hub.layers import ReversibleEmbedding
+from paz.models.transformers.embeddings.reversible import ReversibleEmbedding
 
 from paz.models.transformers import cache as kv_cache
 from paz.models.transformers.logits import soft_cap as apply_soft_cap
--- a/paz/models/foundation/whisper/model.py
+++ b/paz/models/foundation/whisper/model.py
@@ -7,7 +7,7 @@ from keras.activations import gelu
 from keras.layers import Input, Conv1D, Dropout, Add, Lambda
 from keras.layers import LayerNormalization
 from keras.utils import get_file
-from keras_hub.layers import ReversibleEmbedding
+from paz.models.transformers.embeddings.reversible import ReversibleEmbedding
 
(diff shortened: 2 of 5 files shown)


oarriaga:
### Additional validation: GPU + end-to-end model outputs

Ran on an NVIDIA RTX A4000 (JAX `default_backend == gpu`).

**Layer parity vs `keras_hub.layers.ReversibleEmbedding`** (same weights):

| device | tie_weights | forward max|diff| | reverse max|diff| |
|--------|-------------|------------------|------------------|
| GPU | True  | 0.0 | 0.0 |
| GPU | False | 0.0 | 0.0 |
| CPU | True  | 0.0 | 0.0 |
| CPU | False | 0.0 | 0.0 |

**End-to-end model outputs** — built each model twice (internal layer vs
keras_hub), identical weights, compared outputs on GPU. Both exercise the
`reverse=True` logits projection:

- Whisper `WhisperDecoderStep`: logits max abs diff `0.0`, cache `0.0`.
- Gemma4 `Gemma4DecoderStep`: logits max abs diff `0.0`, cache `0.0`.

PR #398, opened 2026-07-10
Add XFeat + LighterGlue feature matching

oarriaga:
Adds a learned local-feature matching subsystem to paz: the XFeat detector-descriptor, the LighterGlue matcher, and a turnkey application, all in the JAX/keras style used by the existing foundation models. Pretrained weights download automatically via `get_file` like the other paz apps.

## What's included

- **`paz.models.XFeat`** — keras backbone (`XFeatModel`) plus a fully jitted extraction pipeline (dense NMS scoring + `top_k`). `backend.py` holds the pure-JAX ops: a sparse `grid_sample` (bicubic/bilinear/nearest, `align_corners=False`) matching torch, dense scoring, and mutual-NN matching.
- **`paz.models.LighterGlue`** — the LightGlue transformer as a subclassed keras `Model` (`SelfBlock`/`CrossBlock`/`FeedForward`/`FourierEncoding`/`MatchAssignment`), mirroring the whisper/dinov2 ports. Match filtering and keypoint normalization stay functional in the wrapper. Runs at full depth (adaptive pruning/early-stop not ported), so it is deterministic and jittable.
- **`paz.applications.MatchXFeat`** — two images in, matched points out (or a side-by-side visualization), downloading both models on first use.
- **`paz.draw.matches`** — draws correspondences across a stacked image pair.
- **`megadepth1500.py`** — a reusable relative-pose AUC evaluation harness.

## Verification (CPU, vs the official implementations)

- **XFeat** on a 480x640 image: keypoints identical, scores max err 1e-7, descriptors 1e-6; mutual-NN matches identical.
- **LighterGlue** on identical keypoints+descriptors: `matches0`/`matches1` identical (123/123); a single self-attention block matches to 1.5e-5; matching scores 1e-4 (the full-depth log-assignment differs by 5e-3 purely from float32 accumulation across layers).
- **Speed** at VGA: XFeat GPU 398 vs 359 FPS (torch), CPU 57 vs 30 FPS (torch). Matcher latency at 2048x2048: GPU 12 ms, CPU 313 ms (torch CPU 233 ms).
- Weights round-trip exact after save/load, and the pretrained auto-download path runs end-to-end.

## Tests

`pytest paz/models/feature paz/applications/feature_matchers_test.py` → all passing. The numerical checks against the reference implementations are opt-in (gated on local reference weights/checkout); the always-on tests cover ops, shapes, and match filtering.


paz/models/feature/lightglue/model.py
```diff
@@ -0,0 +1,228 @@
...
+        keypoints0, descriptors0, keypoints1, descriptors1 = inputs
+        cos0, sin0 = self.encoding(keypoints0)
+        cos1, sin1 = self.encoding(keypoints1)
+        x0 = self.input_proj(descriptors0)
+        x1 = self.input_proj(descriptors1)
+        for self_block, cross_block in zip(self.self_blocks, self.cross_blocks):
+            x0 = self_block(x0, cos0, sin0)
+            x1 = self_block(x1, cos1, sin1)
+            x0, x1 = cross_block(x0, x1)
+        return self.assignment(x0, x1)
```
oarriaga: We should use the Keras Function model API that does not inherit the Model as a class but that constructs the model tree explicitly https://keras.io/guides/functional_api/

paz/models/feature/lightglue/model.py
```diff
@@ -0,0 +1,228 @@
...
+    model = LighterGlueModel(weights)
+    forward = jax.jit(lambda inputs: model(inputs))
+    prune = jax.jit(filter_matches, static_argnums=(1,))
+
+    def call(keypoints0, descriptors0, keypoints1, descriptors1, size0, size1):
+        inputs = [normalize(keypoints0, size0)[None], descriptors0[None],
+                  normalize(keypoints1, size1)[None], descriptors1[None]]
+        scores = forward(inputs)[0]
+        return Matches(*(np.asarray(x) for x in prune(scores,
+                                                      filter_threshold)))
```
oarriaga: This looks messy, it breaks the rule of breaking computation with parenthesis or as list. All computation should happen in a single line.

paz/models/feature/lightglue/model.py
```diff
@@ -0,0 +1,228 @@
...
+
+    return call
+
+
+def LighterGlueModel(weights="pretrained", name="lighterglue"):
+    model = LightGlueTransformer(name=name)
+    build_model(model)
+    if weights == "pretrained":
+        asset = WEIGHTS_URL.rsplit("/", 1)[-1]
+        weights = get_file(asset, WEIGHTS_URL,
```
oarriaga: No parenthesis break. All computation in one line.

paz/models/feature/lightglue/model.py
```diff
@@ -0,0 +1,228 @@
...
+        cos1, sin1 = self.encoding(keypoints1)
+        x0 = self.input_proj(descriptors0)
+        x1 = self.input_proj(descriptors1)
+        for self_block, cross_block in zip(self.self_blocks, self.cross_blocks):
+            x0 = self_block(x0, cos0, sin0)
+            x1 = self_block(x1, cos1, sin1)
+            x0, x1 = cross_block(x0, x1)
+        return self.assignment(x0, x1)
+
+
```
oarriaga: There seems to be many layers here. We should avoid as much as possible making new layers. Instead we should just use functions that represent (wrap) semantically a layer. Layers cause issues because require specifying explicitly how to serialize them. Only in special cases we should add new layers, and if we do they should be completely serializable within Keras. Look at the UNET, SSD model for examples

oarriaga: Apply this for all layers in this PR.

paz/models/feature/lightglue/model.py
```diff
@@ -0,0 +1,228 @@
...
+        value0 = split_heads(self.to_v(x0), NUM_HEADS)
+        value1 = split_heads(self.to_v(x1), NUM_HEADS)
+        similarity = ops.einsum("bhid,bhjd->bhij", query0, query1)
+        message0 = ops.softmax(similarity, axis=-1) @ value1
+        message1 = ops.softmax(swap(similarity), axis=-1) @ value0
+        return self.update(x0, message0), self.update(x1, message1)
+
+    def update(self, x, message):
+        message = self.out_proj(merge_heads(message))
+        return x + self.ffn(ops.concatenate([x, message], axis=-1))
```
oarriaga: We should try to reuse as much as possible the transformers module we have within paz.models.transformers.
Reuse as much computations and functions as possible. If there are elements that could be abstracted into that module we should do that.

oarriaga: Apply that for all the functions in this PR.

paz/models/feature/xfeat/model.py
```diff
@@ -0,0 +1,174 @@
...
+    return Lambda(call, output_shape=(None, None, 64))([target, source])
+
+
+def build_block(x, specs, prefix):
+    for index, (filters, kernel, stride) in enumerate(specs):
+        x = basic_layer(x, filters, kernel, stride, f"{prefix}_{index}")
+    return x
+
+
+def basic_layer(x, filters, kernel, stride, name):
```
oarriaga: This is a bad function name. We should not be as generic as that. Myabe xfeat_conv or something similar.

paz/models/feature/xfeat/model.py
```diff
@@ -0,0 +1,174 @@
+from collections import namedtuple
+
+import numpy as np
+import jax
+import jax.numpy as jp
+import keras
+from keras import ops
+from keras.layers import Input, Conv2D, ReLU, ZeroPadding2D
+from keras.layers import AveragePooling2D, BatchNormalization, Lambda
```
oarriaga: We should avoid using Lambda layers.

paz/models/feature/xfeat/model.py
```diff
@@ -0,0 +1,174 @@
...
+
+def build_heatmap_head(x):
+    x = basic_layer(x, 64, 1, 1, "heatmap_0")
+    x = basic_layer(x, 64, 1, 1, "heatmap_1")
+    x = Conv2D(1, 1, use_bias=True, name="heatmap_out")(x)
+    return keras.activations.sigmoid(x)
+
+
+def build_keypoint_head(normed):
+    x = Lambda(unfold_grid, output_shape=(None, None, 64))(normed)
```
oarriaga: We should avoid using Lambda layers.

oarriaga:
Addressed the review across the whole PR in `eec97fc7`. Summary of the changes, grouped by theme since most comments applied throughout:

**Functional API instead of a subclassed Model** — LighterGlue is now built with the Keras functional API (`build_lighterglue` constructs the graph explicitly and returns a `Model`), no `keras.Model` subclass.

**No custom Layer subclasses** — the `SelfBlock`/`CrossBlock`/`FeedForward`/`FourierEncoding`/`MatchAssignment` classes are gone. Each is now a function that creates and applies plain `Dense`/`LayerNormalization` layers, following the whisper `encoder_block` style. Because the two towers share weights, each block function creates its layers once and applies them to both towers (the layer instances are shared, so weights are shared). This is fully serializable — the model round-trips through `.weights.h5`.

**No Lambda layers** (XFeat) — grayscale, instance-norm and sigmoid use `ops` directly on the symbolic tensors; the pyramid fusion uses `UpSampling2D`; and the keypoint unfold uses `ops.image.extract_patches`. No `Lambda` remains in either model.

**Reuse of the transformers primitives** — the rotary and qkv splits reuse the serializable `SplitDim`/`MergeDims` helpers from `paz.layers` (the same ones `embeddings/rotary.py` uses). For the attention itself, the generic MHA in `paz.models.transformers.attention` assumes separate query/key/value projections and no rotary, whereas LighterGlue uses a fused qkv, a shared query/key projection in cross-attention, and a learnable Fourier rotary applied between projection and scoring — so the attention math stays as small functions here rather than forcing a poor fit. Happy to promote the rotary/dual-softmax helpers into the transformers module if you'd prefer them shared.

**Naming** — `basic_layer` -> `xfeat_layer`, `convolve` -> `xfeat_conv`.

**Single-line computations** — removed the parenthesis/list-broken statements.

Weights were re-ported to the new layer names and re-hosted. Everything stays bit-for-bit: XFeat feats 1.5e-5 / kpts 1.8e-5, LighterGlue matches identical to the reference (123/123). `pytest paz/models/feature paz/applications/feature_matchers_test.py` -> 12 passed.

paz/models/feature/lightglue/port_weights.py
```diff
@@ -0,0 +1,70 @@
...
+    state = matcher_state(torch.load(torch_path, map_location="cpu"))
+    state = {key: value.numpy() for key, value in state.items()}
+    model = LighterGlueModel(weights=None)
+    set_weights(model, state)
+    return model
+
+
+def matcher_state(state):
+    return {key[len("matcher."):]: value for key, value in state.items()
+            if key.startswith("matcher.")}
```
oarriaga: This does not fit in one line we should just have a normal for loop.

paz/models/feature/xfeat/backend.py
```diff
@@ -0,0 +1,147 @@
...
+    top_left = gather_at(features, left, top)
+    top_right = gather_at(features, left + 1, top)
+    bottom_left = gather_at(features, left, top + 1)
+    bottom_right = gather_at(features, left + 1, top + 1)
+    top_row = top_left * (1 - tx) + top_right * tx
+    bottom_row = bottom_left * (1 - tx) + bottom_right * tx
+    return top_row * (1 - ty) + bottom_row * ty
+
+
+def cubic_near(t):
```
oarriaga: This function name is not clear. Functions should start with a verb or "to". Apply this to all functions on this file.

paz/models/feature/xfeat/backend.py
```diff
@@ -0,0 +1,147 @@
...
+    return ((CUBIC_A + 2) * t - (CUBIC_A + 3)) * t * t + 1
+
+
+def cubic_far(t):
```
oarriaga: This function name is not clear. Functions should start with a verb or "to". Apply this to all functions on this file.

paz/models/feature/xfeat/backend.py
```diff
@@ -0,0 +1,147 @@
...
+    return ((CUBIC_A * t - 5 * CUBIC_A) * t + 8 * CUBIC_A) * t - 4 * CUBIC_A
+
+
+def cubic_weights(t):
```
oarriaga: This function name is not clear. Functions should start with a verb or "to". Apply this to all functions on this file.

paz/models/feature/xfeat/backend.py
```diff
@@ -0,0 +1,147 @@
...
+
+def accumulate_row(features, left, rows, weights_x):
+    row = jp.zeros((left.shape[0], features.shape[-1]), features.dtype)
+    for column_offset in range(-1, 3):
+        taps = gather_at(features, left + column_offset, rows)
+        row += taps * weights_x[:, column_offset + 1][:, None]
+    return row
+
+
+def keypoint_heatmap(logits):
```
oarriaga: This function name is not clear. Functions should start with a verb or "to". Apply this to all functions on this file.

paz/models/feature/xfeat/backend.py
```diff
@@ -0,0 +1,147 @@
...
+    scores = jax.nn.softmax(logits, axis=-1)[..., :CELL * CELL]
+    height, width = scores.shape[0], scores.shape[1]
+    scores = jp.reshape(scores, (height, width, CELL, CELL))
+    scores = jp.transpose(scores, (0, 2, 1, 3))
+    return jp.reshape(scores, (height * CELL, width * CELL))
+
+
+def local_maxima(heatmap, kernel):
```
oarriaga: This function name is not clear. Functions should start with a verb or "to". Apply this to all functions on this file.

paz/models/feature/xfeat/backend.py
```diff
@@ -0,0 +1,147 @@
...
+        [(pad, pad), (pad, pad)])
+    return heatmap == pooled
+
+
+def l2_normalize(x, axis=-1):
+    norm = jp.sqrt(jp.sum(x * x, axis=axis, keepdims=True))
+    return x / jp.clip(norm, 1e-12, None)
+
+
+def mutual_nearest_neighbors(descriptors1, descriptors2, min_cosine=-1.0):
```
oarriaga: This function name is not clear. Functions should start with a verb or "to". Apply this to all functions on this file.

paz/models/feature/xfeat/backend.py
```diff
@@ -0,0 +1,147 @@
...
+    mutual = match21[match12] == source
+    if min_cosine > 0:
+        best = jp.max(similarity, axis=1)
+        mutual = mutual & (best > min_cosine)
+    source = np.asarray(source)[np.asarray(mutual)]
+    target = np.asarray(match12)[np.asarray(mutual)]
+    return source, target
+
+
+def full_grid(height, width):
```
oarriaga: This function name is not clear. Functions should start with a verb or "to". Apply this to all functions on this file.

paz/models/feature/xfeat/backend.py
```diff
@@ -0,0 +1,147 @@
...
+    columns, rows = jp.meshgrid(jp.arange(width), jp.arange(height))
+    grid = jp.stack([columns.reshape(-1), rows.reshape(-1)], axis=-1)
+    return grid.astype(jp.float32)
+
+
+def dense_scores(heatmap, heat, threshold, height, width):
```
oarriaga: This function name is not clear. Functions should start with a verb or "to". Apply this to all functions on this file.

paz/models/feature/lightglue/model.py
```diff
@@ -0,0 +1,196 @@
...
+from keras import Input, Model, ops
+from keras.layers import Dense, LayerNormalization
+from keras.utils import get_file
+
+from paz.layers import SplitDim, MergeDims
+
+INPUT_DIM = 64
+DESCRIPTOR_DIM = 96
+HEAD_DIM = 96
+NUM_LAYERS = 6
```
oarriaga: This INPUT_DIM, DESCRIPTOR_DIM, HEAD_DIM, NUM_LAYERS should not be global variables but rather be put where they are used explicitly or as function arguments.

paz/models/feature/lightglue/model.py
```diff
@@ -0,0 +1,196 @@
...
+WEIGHTS_URL = "https://github.com/oarriaga/altamira-data/releases/download/v0.25/xfeat_lighterglue_paz_jax.weights.h5"  # fmt: skip
+
+Matches = namedtuple("Matches", ["matches0", "matches1", "scores0", "scores1"])
+
+
+def LighterGlue(weights="pretrained", filter_threshold=0.1):
+    model = LighterGlueModel(weights)
+    forward = jax.jit(lambda inputs: model(inputs))
```
oarriaga: Do we need to wrap this in a lambda function? It seems that we could maybe just directly use jax.jit(model(inputs))

paz/models/feature/lightglue/model.py
```diff
@@ -0,0 +1,196 @@
...
+        second = prepare(keypoints1, descriptors1, size1)
+        scores = forward(first + second)[0]
+        outputs = prune(scores, filter_threshold)
+        return Matches(*(np.asarray(value) for value in outputs))
+
+    return call
+
+
+def prepare(keypoints, descriptors, size):
+    return [normalize(keypoints, size)[None], descriptors[None]]
```
oarriaga: Most likely the number of keypoints /descriptors is changing at every frame, if they are we should still try to jit normalize.

paz/models/feature/lightglue/model.py
```diff
@@ -0,0 +1,196 @@
...
+    model = LighterGlueModel(weights)
+    forward = jax.jit(lambda inputs: model(inputs))
+    prune = jax.jit(filter_matches, static_argnums=(1,))
+
+    def call(keypoints0, descriptors0, keypoints1, descriptors1, size0, size1):
+        first = prepare(keypoints0, descriptors0, size0)
+        second = prepare(keypoints1, descriptors1, size1)
+        scores = forward(first + second)[0]
+        outputs = prune(scores, filter_threshold)
+        return Matches(*(np.asarray(value) for value in outputs))
```
oarriaga: Can we not jit the whole "call"? If we are always obtaining a variable number of keypoints / descriptors then we can't jit the whole call. But most likely if we are getting a static number of keypoints and descriptors we could jit the whole step and mask where needed. If this is true let's compare both elements and measure their total call time.

paz/models/feature/xfeat/megadepth1500.py
```diff
@@ -0,0 +1,96 @@
+import os
+import json
+import numpy as np
+import cv2
+
```
oarriaga: Why do we need this file?

paz/models/feature/xfeat/model.py
```diff
@@ -0,0 +1,141 @@
...
+def XFeatModel(weights="pretrained", name="xfeat"):
+    model = build_xfeat(name)
+    if weights == "pretrained":
+        asset = WEIGHTS_URL.rsplit("/", 1)[-1]
+        path = get_file(asset, WEIGHTS_URL, cache_subdir="paz/models/xfeat")
+        model.load_weights(path)
+    return model
+
+
+def build_xfeat(name):
```
oarriaga: Let's inline the model building in the XFeatModel function and instead make a function to load the weights.

Do the same for the LighterGlue model.

paz/models/feature/xfeat/model.py
```diff
@@ -0,0 +1,141 @@
...
+def XFeat(weights="pretrained", top_k=4096, threshold=0.05):
+    model = XFeatModel(weights)
+    forward = jax.jit(model)
+    core = jax.jit(extract_core, static_argnums=(3, 4, 5, 6))
+
+    def call(image):
+        tensor, scale = preprocess(image)
+        height, width = tensor.shape[1], tensor.shape[2]
+        outputs = core(*forward(tensor), height, width, top_k, threshold)
+        return finalize(outputs, scale)
```
oarriaga: Can we not simply jit the whole call and mask accordingly? Maybe that would improve performance. Validate that, while keeping the same output contract.

paz/models/feature/lightglue/model.py
oarriaga: Let's keep the same output contract.

oarriaga:
Addressed the new round in `7aeb478f`, including the ones that applied in several places:

**backend.py function names** — renamed every non-verb helper to start with a verb: `compute_cubic_near`/`compute_cubic_far`/`compute_cubic_weights`, `compute_keypoint_heatmap`, `find_local_maxima`, `find_mutual_nearest_neighbors`, `build_pixel_grid`, `compute_dense_scores`, and updated all callers.

**Cramped comprehension → for loop** — `matcher_state` is now a plain for loop.

**LighterGlue globals** — `INPUT_DIM`/`DESCRIPTOR_DIM`/`HEAD_DIM`/`NUM_LAYERS` are no longer module globals; the dims are locals in `LighterGlueModel` and pass down as function arguments (helpers derive head_dim from the tensor shape).

**Inline building + separate load function** — the graph is now constructed directly inside `XFeatModel` and `LighterGlueModel`, and weight downloading moved into a `load_weights` function. Applied to both models.

**Drop the lambda around jit / jit normalize / jit the whole call** — folded preprocessing, keypoint normalization and match filtering into a single jitted step per model, with a named inner function instead of a lambda. I validated total call time both ways, keeping the same output contract; the whole-call version is faster:

| | GPU split | GPU whole | CPU split | CPU whole |
|---|---|---|---|---|
| XFeat | 3.07 ms | **2.47 ms** | 18.1 ms | **16.5 ms** |
| LighterGlue | 38.3 ms | **33.4 ms** | 1387 ms | **1376 ms** |

(top_k 4096; the keypoint/descriptor count does vary per frame, so the jitted step recompiles per size as before — but the whole-call form is still the faster of the two.)

**megadepth1500.py** — removed; the evaluation harness can come back in a dedicated PR.

Still bit-for-bit on CPU (XFeat feats 1.5e-5, LighterGlue matches identical 123/123). `pytest paz/models/feature paz/applications/feature_matchers_test.py` -> 12 passed.

oarriaga:
Follow-up on the jit discussion: implemented the fixed-size + mask design in `99a4f22e`.

The matcher was keying its jit on the keypoint counts `(M, N)`, which change nearly every frame, so it recompiled (~6-7 s) on almost every call. Now both towers are padded to a fixed `capacity` and carry a validity mask; the jitted step always sees the same shapes and compiles once. Structure is a single jitted block (normalize + transformer + assignment + filter, with padded points masked out of attention/assignment) plus a thin eager tail (pad in, slice the fixed-size match arrays back to the real counts out).

Result — latency is now constant across keypoint counts instead of recompiling per size:

| keypoints (M, N) | before | after |
|---|---|---|
| 512 / 492 | ~6-7 s (recompile) | **36 ms** |
| 337 / 317 | ~6-7 s (recompile) | **36 ms** |
| 941 / 921 | ~6-7 s (recompile) | **37 ms** |
| 1500 / 1480 | ~6-7 s (recompile) | **37 ms** |

(GPU, capacity 4096; CPU is a constant ~1.4 s at that capacity.) Masking is bit-for-bit: on CPU the matches are identical to the reference (123/123), scores 1e-4. Masks are inputs, not weights.

Note: at a fixed capacity the matcher always does `capacity²` work, so on CPU with sparse frames there is some wasted compute. Padding to the nearest of a few buckets (as upstream LightGlue does) would trim that; it reuses this same masking and can be a follow-up if needed.

`pytest paz/models/feature paz/applications/feature_matchers_test.py` -> 12 passed.

paz/models/feature/lightglue/model.py
```diff
@@ -0,0 +1,228 @@
...
+
+Matches = namedtuple("Matches", ["matches0", "matches1", "scores0", "scores1"])
+
+
+def LighterGlue(weights="pretrained", filter_threshold=0.1, capacity=4096):
+    model = LighterGlueModel(weights)
+
+    @jax.jit
+    def match(keypoints0, descriptors0, mask0, keypoints1, descriptors1, mask1,
+              size0, size1):
```
oarriaga: Let's always leave an underscore between the name and the number i.e. keypoints_0, etc. Apply this for the whole PR.

paz/models/feature/lightglue/model.py
```diff
@@ -0,0 +1,228 @@
...
+        scores = model(first + second)[0]
+        return filter_matches(scores, mask0, mask1, filter_threshold)
+
+    def call(keypoints0, descriptors0, keypoints1, descriptors1, size0, size1):
+        count0 = min(len(keypoints0), capacity)
+        count1 = min(len(keypoints1), capacity)
+        first = pad_features(keypoints0, descriptors0, capacity)
+        second = pad_features(keypoints1, descriptors1, capacity)
+        outputs = match(*first, *second, size0, size1)
+        m0, m1, s0, s1 = (np.asarray(value) for value in outputs)
```
oarriaga: m and s are not full English variable names.

paz/models/feature/lightglue/model.py
```diff
@@ -0,0 +1,228 @@
...
+    keypoints, descriptors = np.asarray(keypoints), np.asarray(descriptors)
+    count = min(len(keypoints), capacity)
+    mask = np.zeros(capacity, np.float32)
+    mask[:count] = 1.0
+    padded_keypoints = np.zeros((capacity, 2), np.float32)
+    padded_keypoints[:count] = keypoints[:count]
+    padded_descriptors = np.zeros((capacity, descriptors.shape[1]), np.float32)
+    padded_descriptors[:count] = descriptors[:count]
+    return jp.asarray(padded_keypoints), jp.asarray(padded_descriptors), \
+        jp.asarray(mask)
```
oarriaga: This is not following the contract in paz meaning all models should take as input a jax array, and they should also output jax arrays. The drawing function should do the verfication and conversion to the CPU while remaining optimal.

paz/models/feature/lightglue/port_weights.py
```diff
@@ -0,0 +1,75 @@
+from paz.models.feature.lightglue.model import LighterGlueModel
+
+NUM_LAYERS = 6
```
oarriaga: This global variable should be inlined where is used.

paz/models/feature/xfeat/backend.py
```diff
@@ -0,0 +1,147 @@
+import numpy as np
+import jax
+import jax.numpy as jp
+
+CUBIC_A = -0.75
+CELL = 8
```
oarriaga: This global variables should be inlined where they are used.

paz/models/feature/xfeat/backend.py
```diff
@@ -0,0 +1,147 @@
+import numpy as np
+import jax
+import jax.numpy as jp
+
```
oarriaga, line 3: Is there no backend module where this functions could be abstracted for later reused in similar models? For that to be possible, maybe they should be split across multiple backends. Most of these functions seem to operate on "features" we either double down on features as a new backend or we incorporate them in an existing module such as points2D or something similar.

oarriaga:
Addressed all six in `ac73432d`:

- **Underscore before index numbers** — renamed index-suffixed variables across the PR (`keypoints0` -> `keypoints_0`, `x1` -> `x_1`, etc.). Left the weight-key layer-name strings (`block1_0_conv`, `self_attention_0_*`) untouched so the hosted weights still load.
- **`m` / `s`** — replaced with full names (`matches_0`, `scores_0`, ...).
- **jax in / jax out contract** — `XFeat` and `LighterGlue` now keep everything on device and return jax arrays; `paz.draw.matches` does the verification and CPU conversion via `paz.to_numpy`. `MatchXFeat`'s raw output is a `jax.Array`.
- **`NUM_LAYERS` global** — inlined (removed from port_weights).
- **`CUBIC_A` / `CELL` globals** — inlined where used.
- **Backend abstraction** — added `paz.backend.features` (exposed as `paz.features`) holding the reusable local-feature ops: sparse `grid_sample` sampling, `l2_normalize`, `find_mutual_nearest_neighbors`, `build_pixel_grid`, `find_local_maxima`. Future extractors/matchers (ALIKED, DISK, SuperPoint, ...) can reuse these. The xfeat backend keeps only the XFeat-specific head logic (`compute_keypoint_heatmap`, `compute_dense_scores`), which now calls `paz.features`.

Weight layer names are unchanged, so no re-upload was needed. Still bit-for-bit on CPU. `pytest paz/models/feature paz/applications/feature_matchers_test.py` -> 12 passed.

paz/backend/features.py
```diff
@@ -0,0 +1,125 @@
+"""Reusable local-feature ops: sparse sampling, matching, normalization."""
+import numpy as np
+import jax
+import jax.numpy as jp
+
+
+def sample_features(features, positions, height, width, mode):
```
oarriaga: We should not call it sample_features since it would be explicit from paz.features.sample_features (we re repeating the name features twice) it should only sample. Moreover, I am not sure "sample" is the right verb since generally sample is reserved for sampling from a distribution. We should use a better function name.

paz/backend/features.py
```diff
@@ -0,0 +1,125 @@
+"""Reusable local-feature ops: sparse sampling, matching, normalization."""
```
oarriaga: Give a better overview of what "features" constitute. What are they shapes, etc.

paz/models/feature/lightglue/model.py
```diff
@@ -0,0 +1,229 @@
...
+    def call(keypoints_0, descriptors_0, keypoints_1, descriptors_1,
+             size_0, size_1):
+        count_0 = min(len(keypoints_0), capacity)
+        count_1 = min(len(keypoints_1), capacity)
+        first = pad_features(keypoints_0, descriptors_0, capacity)
+        second = pad_features(keypoints_1, descriptors_1, capacity)
+        outputs = match(*first, *second, size_0, size_1)
+        matches_0, matches_1, scores_0, scores_1 = outputs
+        return Matches(matches_0[:count_0], matches_1[:count_1],
+                       scores_0[:count_0], scores_1[:count_1])
```
oarriaga: We should do the :count_0 and :count_1 outside of the Matches wrap. You could split it in two lines namely one for count_0 variables and one for count_1 variables and then a third line for Matches.

paz/models/feature/lightglue/model.py
```diff
@@ -0,0 +1,229 @@
...
+    keypoints_0, descriptors_0 = Input((None, 2)), Input((None, input_dim))
+    keypoints_1, descriptors_1 = Input((None, 2)), Input((None, input_dim))
+    mask_0, mask_1 = Input((None,)), Input((None,))
+    encode = build_encoder(dim, "encoding")
+    project = Dense(dim, name="input_projection")
+    cosine_0, sine_0 = encode(keypoints_0)
+    cosine_1, sine_1 = encode(keypoints_1)
+    x_0, x_1 = project(descriptors_0), project(descriptors_1)
+    for index in range(num_layers):
+        x_0, x_1 = transformer_layer(x_0, x_1, cosine_0, sine_0, cosine_1, sine_1, mask_0, mask_1, dim, index)  # fmt: skip
```
oarriaga: Can you not do a partial of the transformer_layer using *args i.e.
transfomer_args = (cos_0, sin_0, cos_1, sin_1, mask_0, mask_1, dim)
trasnform = partial(transformer_layer, *transformer_args)
Then apply the transform in the for loop?

paz/models/feature/lightglue/model.py
```diff
@@ -0,0 +1,229 @@
...
+    input_dim, dim, num_layers = 64, 96, 6
+    keypoints_0, descriptors_0 = Input((None, 2)), Input((None, input_dim))
+    keypoints_1, descriptors_1 = Input((None, 2)), Input((None, input_dim))
+    mask_0, mask_1 = Input((None,)), Input((None,))
+    encode = build_encoder(dim, "encoding")
+    project = Dense(dim, name="input_projection")
+    cosine_0, sine_0 = encode(keypoints_0)
+    cosine_1, sine_1 = encode(keypoints_1)
+    x_0, x_1 = project(descriptors_0), project(descriptors_1)
+    for index in range(num_layers):
```
oarriaga: Use arg instead of index. Is a paz convention.

paz/models/feature/lightglue/model.py
```diff
@@ -0,0 +1,229 @@
...
+
+    def encode(keypoints):
+        angles = ops.repeat(project(keypoints), 2, axis=-1)
+        return ops.cos(angles), ops.sin(angles)
+
+    return encode
+
+
+def transformer_layer(x_0, x_1, cosine_0, sine_0, cosine_1, sine_1, mask_0,
+                      mask_1, dim, index):
```
oarriaga: It's fine if we use cos, sin instead of the whole name cosine, sine.

oarriaga:
Addressed all six in `0f1fe1ac`:

- **`sample_features` name** — renamed to `paz.features.interpolate` (drops the repeated "features", and avoids "sample" which reads as sampling from a distribution). The pixel-coordinate dispatcher is now `interpolate_at`.
- **`features.py` docstring** — expanded to describe what a feature map is (`(height, width, channels)`) and the position convention (`(N, 2)` of `(x, y)` pixel coordinates).
- **Slicing outside the `Matches` wrap** — split into a `count_0` line, a `count_1` line, then the `Matches(...)` return.
- **`partial` for the transformer stack** — the per-layer-constant args are bundled and `transform = partial(transformer_layer, *arguments)`; the loop just does `x_0, x_1 = transform(arg, x_0, x_1)`.
- **`arg` instead of `index`** — applied to the loop counters across the model and port scripts.
- **`cos` / `sin`** — used instead of `cosine` / `sine`.

Weight layer names are unchanged, so no re-upload. Still bit-for-bit on CPU; `pytest paz/models/feature paz/applications/feature_matchers_test.py` -> 12 passed.

paz/models/feature/xfeat/model.py
```diff
@@ -0,0 +1,145 @@
...
+    positions = grid[chosen]
+    descriptors = features.interpolate(feature_map, positions, height,
+                                       width, "bicubic")
+    return positions, scores, features.l2_normalize(descriptors, axis=-1)
+
+
+def finalize(positions, scores, descriptors, scale):
+    valid = scores > 0
+    keypoints = positions[valid] * scale
+    return Features(keypoints, scores[valid], descriptors[valid])
```
oarriaga: finalize is not a good function name.

PR #400, opened 2026-07-13
Add fused-QKV self-attention primitives to transformers.attention

oarriaga:
## What

Milestone 2 of the canonical DINOv2 / Depth Anything 3 project. Adds general,
reusable fused-QKV self-attention functions to `paz.models.transformers.attention`:

- `project_query_key_value(tokens, hidden_size, use_bias, name)` — single fused
  Dense producing `3 * hidden_size`.
- `split_query_key_value(fused, num_heads, head_dim)` — returns `(query, key,
  value)`, each `(batch, num_heads, seq, head_dim)`.
- `compute_attention(query, key, value)` — scaled dot-product attention from
  already-projected Q, K, V.
- `merge_attention_heads(context)` — `(batch, heads, seq, head_dim)` →
  `(batch, seq, heads*head_dim)`.

These are the exact primitives canonical DINOv2 self-attention composes, and the
shared DA3 base attention will reuse the same project/split/merge (inserting its
own steps between split and `compute_attention`).

## Why these boundaries

The functions are separate, switch-free, and composed explicitly (per the plan:
no boolean-flag mega-attention). The `split_query_key_value` layout reproduces
the parity-proven legacy DINOv2 fused-QKV ordering, so the later weight
converter stays valid.

## Scope decision (deviation from the plan's M2 list)

The plan lists DA3-only primitives (query/key normalization, 2D RoPE,
patch-token position construction) under this milestone. I deferred those to the
DA3 milestone (M4): they are unused by canonical DINOv2 (the immediate next
consumer), and their exact numerical convention (rotate-half vs interleaved
RoPE, qk-norm form, position layout) must be verified against the pinned DA3
reference, which is not yet established. Building them now would be unverifiable
and likely reworked. Added there, tested against the reference, they compose
cleanly on top of these primitives.

## Verification

- `KERAS_BACKEND=jax pytest paz/models/transformers` — **41 passed**.
- New tests cover: fused output width, split shapes, a convention-lock test that
  the fused axis splits as `(3, heads, head_dim)`, a NumPy parity check for the
  attention math, and head-merge shape.
- Functional-model smoke: primitives build a `keras.Model` and JIT output
  matches eager (`allclose`, atol 1e-5).
- Existing `attend`/`kv_attend`/`build_cache`/`kernel` untouched; whisper imports
  and all consumer tests pass. pyflakes clean.

### DIFF
--- a/paz/models/transformers/attention.py
+++ b/paz/models/transformers/attention.py
@@ -2,7 +2,7 @@ import string
 
 import keras
 from keras import ops
-from keras.layers import Dropout, EinsumDense
+from keras.layers import Dense, Dropout, EinsumDense, Reshape
 
 from paz.models.transformers import cache as kv_cache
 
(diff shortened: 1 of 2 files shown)


oarriaga:
Superseded: consolidating milestones 2+ into a single PR (#401) per request.

PR #401, opened 2026-07-13
Canonical DINOv2 + Depth Anything 3 (consolidated)

oarriaga:
Consolidated PR for the canonical DINOv2 / Depth Anything 3 port (milestone 1
merged separately as #399). Every model component is verified bit-for-bit
against the pinned references in the same process (opt-in via DA3_REFERENCE).

## M2 — transformer primitives (`paz/models/transformers`)
Fused-QKV self-attention + DA3 primitives (2D RoPE, per-head QK-norm, patch
positions), verified ~5e-7. Existing attention untouched.

## M3 — canonical function-only DINOv2 (`foundation/dinov2`)
`(class_token, patch_tokens)` / feature-map tuples; canonical == legacy
bit-for-bit; strict converter, save/reload, JIT. Exported from `paz.models`.

## M4/M6 — DA3 any-view SMALL and BASE (`foundation/depth_anything3`)
Backbone (view routing, camera tokens, alternating local/global attention, 2D
RoPE, QK-norm), DualDPT head (depth+rays), camera decoder. `build_da3_small` /
`build_da3_base` return `depth, depth_confidence, extrinsics, intrinsics, rays,
ray_confidence`. BASE reuses the same functions. End-to-end parity vs real
Apache-2.0 checkpoints ~1e-6.

## M7 — permissive monocular/metric large
`build_da3_mono_large` / `build_da3_metric_large`: plain DINOv2-Large backbone +
standard DPT head with sky → `(depth, sky)`. Verified vs DA3MONO-LARGE.

## M5 — weights
`convert.py` (small/base/mono/metric): download → port → save `.weights.h5` →
reload/verify → license+checksum metadata. Weights not committed.

## Applications and examples
`paz.applications.depth_estimators`: `EstimateDepthAnything3Small/Base/MonoLarge/
MetricLarge` (function factories returning closures). Preprocessing runs outside
the compiled model; models build per view count and load a converted weights
file. Examples under `examples/depth_anything3/`.

## Verification (`KERAS_BACKEND=jax`)
- `pytest paz/models/transformers paz/models/foundation/{dinov2,depth_anything3}
  paz/applications/depth_estimators_test.py` — all pass (5 skipped opt-in ref).
- Opt-in reference (`DA3_REFERENCE=1`): DINOv2, DA3 backbone, SMALL e2e, BASE
  e2e, MONO-LARGE all pass. Real-weights app smoke: finite depth, sane cameras.
- No named tuples, tensor dicts, or custom classes. No runtime torch/xformers.

## Remaining
Weight hosting only (maintainer infra); the converter produces files + metadata.

### DIFF
--- a/examples/depth_anything3/estimate_metric.py
+++ b/examples/depth_anything3/estimate_metric.py
@@ -1,6 +1,6 @@
 """Metric monocular depth in meters with Depth Anything 3.
 
-Convert the Apache-2.0 DA3METRIC-LARGE checkpoint first:
+Convert the DA3METRIC-LARGE checkpoint first:
 
     DA3_MODEL=metric python -m paz.models.foundation.depth_anything3.convert
 
--- a/examples/depth_anything3/estimate_metric.py
+++ b/examples/depth_anything3/estimate_metric.py
@@ -9,9 +9,9 @@ explicitly; it is never inferred.
 """
 import argparse
 
-import cv2
 import numpy as np
 
+import paz
 from paz.applications import EstimateDepthAnything3MetricLarge
 
 if __name__ == "__main__":
@@ -22,12 +22,12 @@ if __name__ == "__main__":
     parser.add_argument("--image_size", type=int, default=518)
     args = parser.parse_args()
 
-    estimate = EstimateDepthAnything3MetricLarge(
-        args.weights, args.focal_length, args.image_size)
-    image = cv2.cvtColor(cv2.imread(args.image), cv2.COLOR_BGR2RGB)
-    depth_meters, sky = estimate(image)
+    settings = args.weights, args.focal_length, args.image_size
+    estimate = EstimateDepthAnything3MetricLarge(*settings)
+    depth_meters, sky = estimate(paz.image.load(args.image))
 
     depth_meters = np.array(depth_meters)[0]
+    lowest = round(float(depth_meters.min()), 3)
+    highest = round(float(depth_meters.max()), 3)
     print("median depth (m):", round(float(np.median(depth_meters)), 3))
-    print("range (m):", round(float(depth_meters.min()), 3),
-          round(float(depth_meters.max()), 3))
+    print("range (m):", lowest, highest)
--- a/examples/depth_anything3/estimate_metric.py
+++ b/examples/depth_anything3/estimate_metric.py
@@ -1,0 +1,33 @@
+"""Metric monocular depth in meters with Depth Anything 3.
+
+Convert the Apache-2.0 DA3METRIC-LARGE checkpoint first:
+
+    DA3_MODEL=metric python -m paz.models.foundation.depth_anything3.convert
+
+Focal length is in pixels at the processed resolution and must be given
+explicitly; it is never inferred.
+"""
+import argparse
+
+import cv2
+import numpy as np
+
+from paz.applications import EstimateDepthAnything3MetricLarge
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--weights", required=True)
+    parser.add_argument("--image", required=True)
+    parser.add_argument("--focal_length", type=float, required=True)
+    parser.add_argument("--image_size", type=int, default=518)
+    args = parser.parse_args()
+
+    estimate = EstimateDepthAnything3MetricLarge(
+        args.weights, args.focal_length, args.image_size)
+    image = cv2.cvtColor(cv2.imread(args.image), cv2.COLOR_BGR2RGB)
+    depth_meters, sky = estimate(image)
+
+    depth_meters = np.array(depth_meters)[0]
+    print("median depth (m):", round(float(np.median(depth_meters)), 3))
+    print("range (m):", round(float(depth_meters.min()), 3),
+          round(float(depth_meters.max()), 3))
--- a/examples/depth_anything3/estimate_multi_view.py
+++ b/examples/depth_anything3/estimate_multi_view.py
@@ -1,0 +1,30 @@
+"""Any-view depth, rays, and recovered cameras with Depth Anything 3.
+
+Convert the Apache-2.0 DA3-SMALL checkpoint first:
+
+    python -m paz.models.foundation.depth_anything3.convert
+"""
+import argparse
+
+import cv2
+import numpy as np
+
+from paz.applications import EstimateDepthAnything3Small
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--weights", required=True)
+    parser.add_argument("--images", nargs="+", required=True)
+    parser.add_argument("--image_size", type=int, default=518)
+    args = parser.parse_args()
+
+    estimate = EstimateDepthAnything3Small(args.weights, args.image_size)
+    images = [cv2.cvtColor(cv2.imread(path), cv2.COLOR_BGR2RGB)
+              for path in args.images]
+    depth, confidence, extrinsics, intrinsics, rays, ray_confidence = estimate(images)
+
+    print("depth", tuple(depth.shape))
+    for view, (extrinsic, intrinsic) in enumerate(zip(np.array(extrinsics)[0],
+                                                      np.array(intrinsics)[0])):
+        print(f"view {view} focal", round(float(intrinsic[0, 0]), 1),
+              "translation", extrinsic[:, 3].round(3).tolist())
--- a/examples/depth_anything3/estimate_multi_view.py
+++ b/examples/depth_anything3/estimate_multi_view.py
@@ -6,9 +6,9 @@ Convert the Apache-2.0 DA3-SMALL checkpoint first:
 """
 import argparse
 
-import cv2
 import numpy as np
 
+import paz
 from paz.applications import EstimateDepthAnything3Small
 
 if __name__ == "__main__":
@@ -19,12 +19,14 @@ if __name__ == "__main__":
     args = parser.parse_args()
 
     estimate = EstimateDepthAnything3Small(args.weights, args.image_size)
-    images = [cv2.cvtColor(cv2.imread(path), cv2.COLOR_BGR2RGB)
-              for path in args.images]
-    depth, confidence, extrinsics, intrinsics, rays, ray_confidence = estimate(images)
+    images = [paz.image.load(path) for path in args.images]
+    outputs = estimate(images)
+    depth, confidence, extrinsics, intrinsics, rays, ray_confidence = outputs
 
     print("depth", tuple(depth.shape))
-    for view, (extrinsic, intrinsic) in enumerate(zip(np.array(extrinsics)[0],
-                                                      np.array(intrinsics)[0])):
-        print(f"view {view} focal", round(float(intrinsic[0, 0]), 1),
-              "translation", extrinsic[:, 3].round(3).tolist())
+    extrinsics = np.array(extrinsics)[0]
+    intrinsics = np.array(intrinsics)[0]
+    for view in range(len(extrinsics)):
+        focal = round(float(intrinsics[view][0, 0]), 1)
+        translation = extrinsics[view][:, 3].round(3).tolist()
+        print(f"view {view} focal", focal, "translation", translation)
--- a/examples/depth_anything3/estimate_multi_view.py
+++ b/examples/depth_anything3/estimate_multi_view.py
@@ -1,6 +1,6 @@
 """Any-view depth, rays, and recovered cameras with Depth Anything 3.
 
-Convert the Apache-2.0 DA3-SMALL checkpoint first:
+Convert the DA3-SMALL checkpoint first:
 
     python -m paz.models.foundation.depth_anything3.convert
 """
--- a/examples/depth_anything3/estimate_single_view.py
+++ b/examples/depth_anything3/estimate_single_view.py
@@ -6,9 +6,9 @@ Convert the Apache-2.0 DA3MONO-LARGE checkpoint first:
 """
 import argparse
 
-import cv2
 import numpy as np
 
+import paz
 from paz.applications import EstimateDepthAnything3MonoLarge
 
 if __name__ == "__main__":
@@ -20,10 +20,10 @@ if __name__ == "__main__":
     args = parser.parse_args()
 
     estimate = EstimateDepthAnything3MonoLarge(args.weights, args.image_size)
-    image = cv2.cvtColor(cv2.imread(args.image), cv2.COLOR_BGR2RGB)
-    depth, sky = estimate(image)
+    depth, sky = estimate(paz.image.load(args.image))
 
     depth = np.array(depth)[0]
     scaled = (depth - depth.min()) / (depth.max() - depth.min() + 1e-8)
-    cv2.imwrite(args.output, (scaled * 255).astype("uint8"))
+    gray = (scaled[..., None] * np.ones(3) * 255).astype("uint8")
+    paz.image.write(args.output, gray)
     print("saved", args.output)
--- a/examples/depth_anything3/estimate_single_view.py
+++ b/examples/depth_anything3/estimate_single_view.py
@@ -1,0 +1,29 @@
+"""Monocular relative depth with Depth Anything 3.
+
+Convert the Apache-2.0 DA3MONO-LARGE checkpoint first:
+
+    DA3_MODEL=mono python -m paz.models.foundation.depth_anything3.convert
+"""
+import argparse
+
+import cv2
+import numpy as np
+
+from paz.applications import EstimateDepthAnything3MonoLarge
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--weights", required=True)
+    parser.add_argument("--image", required=True)
+    parser.add_argument("--image_size", type=int, default=518)
+    parser.add_argument("--output", default="depth.png")
+    args = parser.parse_args()
+
+    estimate = EstimateDepthAnything3MonoLarge(args.weights, args.image_size)
+    image = cv2.cvtColor(cv2.imread(args.image), cv2.COLOR_BGR2RGB)
+    depth, sky = estimate(image)
+
+    depth = np.array(depth)[0]
+    scaled = (depth - depth.min()) / (depth.max() - depth.min() + 1e-8)
+    cv2.imwrite(args.output, (scaled * 255).astype("uint8"))
+    print("saved", args.output)
--- a/examples/depth_anything3/estimate_single_view.py
+++ b/examples/depth_anything3/estimate_single_view.py
@@ -1,6 +1,6 @@
 """Monocular relative depth with Depth Anything 3.
 
-Convert the Apache-2.0 DA3MONO-LARGE checkpoint first:
+Convert the DA3MONO-LARGE checkpoint first:
 
     DA3_MODEL=mono python -m paz.models.foundation.depth_anything3.convert
 """
--- a/paz/applications/depth_estimators.py
+++ b/paz/applications/depth_estimators.py
@@ -10,73 +10,67 @@ Any-view estimators return, in order:
 Monocular estimators return ``depth, sky``; the metric estimator returns
 ``depth_meters, sky``.
 """
-import cv2
 import numpy as np
 from keras import ops
 
-from paz.models.foundation.depth_anything3.models import build_da3_small
-from paz.models.foundation.depth_anything3.models import build_da3_base
-from paz.models.foundation.depth_anything3.models import build_da3_mono_large
-from paz.models.foundation.depth_anything3.models import build_da3_metric_large
-
-IMAGENET_MEAN = np.array([0.485, 0.456, 0.406], "float32")
-IMAGENET_STD = np.array([0.229, 0.224, 0.225], "float32")
-METRIC_DIVISOR = 300.0
+from paz.backend.image import resize_opencv, standardize
+from paz.models import DepthAnything3Small, DepthAnything3Base
+from paz.models import DepthAnything3MonoLarge, DepthAnything3MetricLarge
 
 
 def EstimateDepthAnything3Small(weights_path, image_size=518):
-    return build_any_view_estimator(build_da3_small, weights_path, image_size)
+    args = DepthAnything3Small, weights_path, image_size
+    return build_any_view_estimator(*args)
 
 
 def EstimateDepthAnything3Base(weights_path, image_size=518):
-    return build_any_view_estimator(build_da3_base, weights_path, image_size)
+    args = DepthAnything3Base, weights_path, image_size
+    return build_any_view_estimator(*args)
 
 
 def build_any_view_estimator(builder, weights_path, image_size):
-    models = {}
+    cache = {}
 
     def estimate(images):
         views = preprocess_views(images, image_size)
-        model = load_for_views(models, builder, views.shape[1], image_size,
-                               weights_path)
-        depth, confidence, extrinsics, intrinsics, rays, ray_confidence = model(views)
-        return depth, confidence, extrinsics, intrinsics, rays, ray_confidence
+        args = cache, builder, views.shape[1], image_size, weights_path
+        return load_view_model(*args)(views)
 
     return estimate
 
 
 def EstimateDepthAnything3MonoLarge(weights_path, image_size=518):
-    model = build_mono_model(build_da3_mono_large, weights_path, image_size)
+    model = load_mono_model(DepthAnything3MonoLarge, weights_path, image_size)
 
     def estimate(image):
-        depth, sky = model(preprocess_batch(image, image_size))
-        return depth, sky
+        return model(preprocess_batch(image, image_size))
 
     return estimate
 
 
-def EstimateDepthAnything3MetricLarge(weights_path, focal_length, image_size=518):
-    model = build_mono_model(build_da3_metric_large, weights_path, image_size)
+def EstimateDepthAnything3MetricLarge(weights_path, focal_length,
+                                      image_size=518):
+    model = load_mono_model(DepthAnything3MetricLarge, weights_path, image_size)
 
     def estimate(image):
         depth, sky = model(preprocess_batch(image, image_size))
-        return focal_length * depth / METRIC_DIVISOR, sky
+        return focal_length * depth / 300.0, sky
 
     return estimate
 
 
-def build_mono_model(builder, weights_path, image_size):
+def load_mono_model(builder, weights_path, image_size):
     model = builder((image_size, image_size, 3))
     model.load_weights(weights_path)
     return model
 
 
-def load_for_views(models, builder, num_views, image_size, weights_path):
-    if num_views not in models:
+def load_view_model(cache, builder, num_views, image_size, weights_path):
+    if num_views not in cache:
         model = builder(num_views, (image_size, image_size, 3))
         model.load_weights(weights_path)
-        models[num_views] = model
-    return models[num_views]
+        cache[num_views] = model
+    return cache[num_views]
 
 
 def preprocess_views(images, image_size):
@@ -89,10 +83,15 @@ def preprocess_batch(image, image_size):
 
 
 def preprocess_image(image, image_size):
+    mean = np.array([0.485, 0.456, 0.406], "float32")
+    std = np.array([0.229, 0.224, 0.225], "float32")
+    array = ops.convert_to_tensor(to_float(image))
+    resized = resize_opencv(array, (image_size, image_size))
+    return standardize(resized, mean, std)
+
+
+def to_float(image):
     image = np.asarray(image)
     if image.dtype == np.uint8:
-        image = image.astype("float32") / 255.0
-    resized = cv2.resize(image.astype("float32"), (image_size, image_size),
-                         interpolation=cv2.INTER_AREA)
-    normalized = (resized - IMAGENET_MEAN) / IMAGENET_STD
-    return ops.convert_to_tensor(normalized, dtype="float32")
+        return image.astype("float32") / 255.0
+    return image.astype("float32")
--- a/paz/applications/depth_estimators.py
+++ b/paz/applications/depth_estimators.py
@@ -1,0 +1,98 @@
+"""High-level Depth Anything 3 estimators.
+
+Preprocessing happens outside the compiled model: images are resized to a
+fixed square resolution (a multiple of the patch size), scaled to [0, 1], and
+ImageNet-normalized. Weights come from a converted ``.weights.h5`` file; pass
+its path explicitly until hosting is available.
+
+Any-view estimators return, in order:
+``depth, depth_confidence, extrinsics, intrinsics, rays, ray_confidence``.
+Monocular estimators return ``depth, sky``; the metric estimator returns
+``depth_meters, sky``.
+"""
+import cv2
+import numpy as np
+from keras import ops
+
+from paz.models.foundation.depth_anything3.models import build_da3_small
+from paz.models.foundation.depth_anything3.models import build_da3_base
+from paz.models.foundation.depth_anything3.models import build_da3_mono_large
+from paz.models.foundation.depth_anything3.models import build_da3_metric_large
+
+IMAGENET_MEAN = np.array([0.485, 0.456, 0.406], "float32")
+IMAGENET_STD = np.array([0.229, 0.224, 0.225], "float32")
+METRIC_DIVISOR = 300.0
+
+
+def EstimateDepthAnything3Small(weights_path, image_size=518):
+    return build_any_view_estimator(build_da3_small, weights_path, image_size)
+
+
+def EstimateDepthAnything3Base(weights_path, image_size=518):
+    return build_any_view_estimator(build_da3_base, weights_path, image_size)
+
+
+def build_any_view_estimator(builder, weights_path, image_size):
+    models = {}
+
+    def estimate(images):
+        views = preprocess_views(images, image_size)
+        model = load_for_views(models, builder, views.shape[1], image_size,
+                               weights_path)
+        depth, confidence, extrinsics, intrinsics, rays, ray_confidence = model(views)
+        return depth, confidence, extrinsics, intrinsics, rays, ray_confidence
+
+    return estimate
+
+
+def EstimateDepthAnything3MonoLarge(weights_path, image_size=518):
+    model = build_mono_model(build_da3_mono_large, weights_path, image_size)
+
+    def estimate(image):
+        depth, sky = model(preprocess_batch(image, image_size))
+        return depth, sky
+
+    return estimate
+
+
+def EstimateDepthAnything3MetricLarge(weights_path, focal_length, image_size=518):
+    model = build_mono_model(build_da3_metric_large, weights_path, image_size)
+
+    def estimate(image):
+        depth, sky = model(preprocess_batch(image, image_size))
+        return focal_length * depth / METRIC_DIVISOR, sky
+
+    return estimate
+
+
+def build_mono_model(builder, weights_path, image_size):
+    model = builder((image_size, image_size, 3))
+    model.load_weights(weights_path)
+    return model
+
+
+def load_for_views(models, builder, num_views, image_size, weights_path):
+    if num_views not in models:
+        model = builder(num_views, (image_size, image_size, 3))
+        model.load_weights(weights_path)
+        models[num_views] = model
+    return models[num_views]
+
+
+def preprocess_views(images, image_size):
+    views = [preprocess_image(image, image_size) for image in images]
+    return ops.expand_dims(ops.stack(views, axis=0), axis=0)
+
+
+def preprocess_batch(image, image_size):
+    return ops.expand_dims(preprocess_image(image, image_size), axis=0)
+
+
+def preprocess_image(image, image_size):
+    image = np.asarray(image)
+    if image.dtype == np.uint8:
+        image = image.astype("float32") / 255.0
+    resized = cv2.resize(image.astype("float32"), (image_size, image_size),
+                         interpolation=cv2.INTER_AREA)
+    normalized = (resized - IMAGENET_MEAN) / IMAGENET_STD
+    return ops.convert_to_tensor(normalized, dtype="float32")
--- a/paz/backend/image.py
+++ b/paz/backend/image.py
@@ -81,6 +81,39 @@ def resize(image, size, method="linear", antialias=False):
     return jax.image.resize(image, (*size, image.shape[-1]), method, antialias)
 
 
+def resize_bilinear_align_corners(image, size):
+    """Bilinear resize matching torch F.interpolate(align_corners=True).
+
+    Operates on channels-last ``(batch, height, width, channels)`` tensors and
+    uses ``keras.ops`` so it traces inside functional models and runs eagerly.
+    Separable: interpolate rows, then columns.
+    """
+    from keras import ops
+    rows = interpolate_bilinear_axis(image, 1, size[0], ops)
+    return interpolate_bilinear_axis(rows, 2, size[1], ops)
+
+
+def interpolate_bilinear_axis(image, axis, target, ops):
+    source = image.shape[axis]
+    lower, upper, weight = bilinear_sample_positions(source, target, ops)
+    low = ops.take(image, lower, axis=axis)
+    high = ops.take(image, upper, axis=axis)
+    shape = [1] * len(image.shape)
+    shape[axis] = target
+    weight = ops.reshape(weight, shape)
+    return low * (1.0 - weight) + high * weight
+
+
+def bilinear_sample_positions(source, target, ops):
+    if target == 1 or source == 1:
+        positions = ops.zeros((target,))
+    else:
+        positions = ops.arange(target) * ((source - 1) / (target - 1))
+    lower = ops.cast(ops.floor(positions), "int32")
+    upper = ops.minimum(lower + 1, source - 1)
+    return lower, upper, positions - lower
+
+
 def resize_opencv(image: jax.Array, size: tuple[int, int]) -> jax.Array:
     # TODO change to split size into H, W
     data = jax.ShapeDtypeStruct((size[0], size[1], image.shape[2]), image.dtype)
--- a/paz/models/foundation/depth_anything3/backbone.py
+++ b/paz/models/foundation/depth_anything3/backbone.py
@@ -1,82 +1,86 @@
 from keras import ops
 from keras.layers import LayerNormalization
 
-from paz.models.foundation.dinov2.blocks import build_dinov2_block
+from paz.models.foundation.dinov2 import blocks as dinov2_blocks
 from paz.models.foundation.depth_anything3 import routing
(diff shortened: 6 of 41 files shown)


paz/applications/depth_estimators.py
```diff
@@ -0,0 +1,98 @@
...
+from keras import ops
+
+from paz.models.foundation.depth_anything3.models import build_da3_small
+from paz.models.foundation.depth_anything3.models import build_da3_base
+from paz.models.foundation.depth_anything3.models import build_da3_mono_large
+from paz.models.foundation.depth_anything3.models import build_da3_metric_large
+
+IMAGENET_MEAN = np.array([0.485, 0.456, 0.406], "float32")
+IMAGENET_STD = np.array([0.229, 0.224, 0.225], "float32")
+METRIC_DIVISOR = 300.0
```
oarriaga: Let's not have these type of global variables. Inline them where needed.

paz/applications/depth_estimators.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+def preprocess_image(image, image_size):
+    image = np.asarray(image)
+    if image.dtype == np.uint8:
+        image = image.astype("float32") / 255.0
+    resized = cv2.resize(image.astype("float32"), (image_size, image_size),
+                         interpolation=cv2.INTER_AREA)
+    normalized = (resized - IMAGENET_MEAN) / IMAGENET_STD
+    return ops.convert_to_tensor(normalized, dtype="float32")
```
oarriaga: Use paz internal backend functions for this i.e. paz.backend.image.

paz/backend/image.py
```diff
@@ -81,6 +81,39 @@ def resize(image, size, method="linear", antialias=False):
...
 
 
+def resize_bilinear_align_corners(image, size):
+    """Bilinear resize matching torch F.interpolate(align_corners=True).
+
+    Operates on channels-last ``(batch, height, width, channels)`` tensors and
+    uses ``keras.ops`` so it traces inside functional models and runs eagerly.
+    Separable: interpolate rows, then columns.
+    """
+    from keras import ops
```
oarriaga, line 91: Don't import a module inside a function.

paz/backend/image.py
```diff
@@ -81,6 +81,39 @@ def resize(image, size, method="linear", antialias=False):
...
+    rows = interpolate_bilinear_axis(image, 1, size[0], ops)
+    return interpolate_bilinear_axis(rows, 2, size[1], ops)
+
+
+def interpolate_bilinear_axis(image, axis, target, ops):
+    source = image.shape[axis]
+    lower, upper, weight = bilinear_sample_positions(source, target, ops)
+    low = ops.take(image, lower, axis=axis)
+    high = ops.take(image, upper, axis=axis)
+    shape = [1] * len(image.shape)
```
oarriaga, line 101: This operation [1] * len(image.shape) it's unclear, refactor for clarity.

paz/backend/image.py
```diff
@@ -81,6 +81,39 @@ def resize(image, size, method="linear", antialias=False):
...
+
+
+def bilinear_sample_positions(source, target, ops):
+    if target == 1 or source == 1:
+        positions = ops.zeros((target,))
+    else:
+        positions = ops.arange(target) * ((source - 1) / (target - 1))
+    lower = ops.cast(ops.floor(positions), "int32")
+    upper = ops.minimum(lower + 1, source - 1)
+    return lower, upper, positions - lower
```
oarriaga, line 114: Functions should do one thing. This function splits different computations based on wether or not target / source == 1? It's also not clear why the variable names are "target" and "source" and those seem to hold ints. The variable name does not match and does not correctly name the contents of the variable..

examples/depth_anything3/estimate_metric.py
```diff
@@ -0,0 +1,33 @@
...
+
+    estimate = EstimateDepthAnything3MetricLarge(
+        args.weights, args.focal_length, args.image_size)
+    image = cv2.cvtColor(cv2.imread(args.image), cv2.COLOR_BGR2RGB)
+    depth_meters, sky = estimate(image)
+
+    depth_meters = np.array(depth_meters)[0]
+    print("median depth (m):", round(float(np.median(depth_meters)), 3))
+    print("range (m):", round(float(depth_meters.min()), 3),
+          round(float(depth_meters.max()), 3))
```
oarriaga: Split computation in multiple lines so that everything fits under 80 columns per line with out line breaks.

examples/depth_anything3/estimate_multi_view.py
```diff
@@ -0,0 +1,30 @@
...
+    estimate = EstimateDepthAnything3Small(args.weights, args.image_size)
+    images = [cv2.cvtColor(cv2.imread(path), cv2.COLOR_BGR2RGB)
+              for path in args.images]
+    depth, confidence, extrinsics, intrinsics, rays, ray_confidence = estimate(images)
+
+    print("depth", tuple(depth.shape))
+    for view, (extrinsic, intrinsic) in enumerate(zip(np.array(extrinsics)[0],
+                                                      np.array(intrinsics)[0])):
+        print(f"view {view} focal", round(float(intrinsic[0, 0]), 1),
+              "translation", extrinsic[:, 3].round(3).tolist())
```
oarriaga: Split computation into multiple lines so that every single line fits under 80 columns and we don't break lines with parenthesis into multiple lines.

examples/depth_anything3/estimate_multi_view.py
```diff
@@ -0,0 +1,30 @@
...
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--weights", required=True)
+    parser.add_argument("--images", nargs="+", required=True)
+    parser.add_argument("--image_size", type=int, default=518)
+    args = parser.parse_args()
+
+    estimate = EstimateDepthAnything3Small(args.weights, args.image_size)
+    images = [cv2.cvtColor(cv2.imread(path), cv2.COLOR_BGR2RGB)
+              for path in args.images]
```
oarriaga: Use paz.image directly. Avoid using openCV directly.

oarriaga: This rule applies through out the whole PR.

examples/depth_anything3/estimate_single_view.py
```diff
@@ -0,0 +1,29 @@
...
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--weights", required=True)
+    parser.add_argument("--image", required=True)
+    parser.add_argument("--image_size", type=int, default=518)
+    parser.add_argument("--output", default="depth.png")
+    args = parser.parse_args()
+
+    estimate = EstimateDepthAnything3MonoLarge(args.weights, args.image_size)
+    image = cv2.cvtColor(cv2.imread(args.image), cv2.COLOR_BGR2RGB)
```
oarriaga: Use paz.image directly. This applies for the whole PR.

paz/applications/depth_estimators.py
```diff
@@ -0,0 +1,98 @@
...
+    return build_any_view_estimator(build_da3_base, weights_path, image_size)
+
+
+def build_any_view_estimator(builder, weights_path, image_size):
+    models = {}
+
+    def estimate(images):
+        views = preprocess_views(images, image_size)
+        model = load_for_views(models, builder, views.shape[1], image_size,
+                               weights_path)
```
oarriaga: Let's try to split computation or pack in args so that everything fits in a single line and it does not break into multiple lines. It's generally easier to read.

paz/models/foundation/depth_anything3/backbone.py
```diff
@@ -0,0 +1,82 @@
+from keras import ops
+from keras.layers import LayerNormalization
+
+from paz.models.foundation.dinov2.blocks import build_dinov2_block
+from paz.models.foundation.depth_anything3 import routing
+from paz.models.foundation.depth_anything3.blocks import build_da3_block
+from paz.models.foundation.depth_anything3.embeddings import insert_camera_tokens
+
+LAYER_NORM_EPSILON = 1e-6
```
oarriaga: No global variables.

paz/models/foundation/depth_anything3/backbone.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+def build_da3_backbone(tokens, num_views, grid, hidden_size, depth, num_heads,
+                       mlp_ratio, layer_scale_init, out_layers, alt_start):
+    positions = build_positions(num_views, grid)
+    collected = run_da3_encoder(tokens, num_views, positions, hidden_size,
+                                depth, num_heads, mlp_ratio, layer_scale_init,
+                                out_layers, alt_start)
```
oarriaga: Try packing into args such that everything fits in a single line. We should try to split large functions like this such that they don't take so many parameters.

paz/models/foundation/depth_anything3/backbone.py
```diff
@@ -0,0 +1,82 @@
...
+from paz.models.foundation.depth_anything3.blocks import build_da3_block
+from paz.models.foundation.depth_anything3.embeddings import insert_camera_tokens
+
+LAYER_NORM_EPSILON = 1e-6
+
+
+def build_da3_backbone(tokens, num_views, grid, hidden_size, depth, num_heads,
+                       mlp_ratio, layer_scale_init, out_layers, alt_start):
+    positions = build_positions(num_views, grid)
+    collected = run_da3_encoder(tokens, num_views, positions, hidden_size,
```
oarriaga: run_da3_encoder is maybe not the best possible function name. It's unclear what "run" means. Moreover, da3 is an acronym for depth anything 3; however, we are already inside the depth anything 3 model module; thus, if that distinction is not necessary we should remove it.

paz/models/foundation/depth_anything3/backbone.py
```diff
@@ -0,0 +1,82 @@
...
+        args = (hidden_size, num_heads, mlp_ratio, layer_scale_init, index)
+        if is_global_block(index, alt_start):
+            tokens = apply_global_block(tokens, num_views, global_positions, *args)
+        else:
+            tokens = apply_local_block(tokens, num_views, local_positions,
+                                       *args, alt_start)
+            local_tokens = tokens
+        if index in out_layers:
+            collected.append(ops.concatenate([local_tokens, tokens], axis=-1))
+    return collected
```
oarriaga: This function is too large. Functions are usually ~6 lines long. Let's try to split it up.

paz/models/foundation/depth_anything3/routing.py
```diff
@@ -0,0 +1,60 @@
...
+Tokens carry an explicit view axis as ``(batch, views, tokens, channels)``.
+Local blocks fold views into the batch so each view attends to itself; global
+blocks merge views into one sequence so all views attend jointly.
+"""
+import numpy as np
+from keras import ops
+
+
+def fold_view_images(images):
+    height, width, channels = images.shape[-3:]
```
oarriaga: Just unpack naturally all the shape i.e. images.shape even though you might not use all variables.

paz/models/foundation/depth_anything3/routing.py
```diff
@@ -0,0 +1,60 @@
...
+    return ops.reshape(images, (-1, height, width, channels))
+
+
+def fold_views_into_batch(tokens):
+    channels = tokens.shape[-1]
```
oarriaga: Just unpack naturally all the shape i.e. images.shape even though you might not use all variables.

paz/models/foundation/depth_anything3/routing.py
```diff
@@ -0,0 +1,60 @@
...
+    num_tokens = tokens.shape[-2]
```
oarriaga: just unpack once.

paz/applications/depth_estimators.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+def build_mono_model(builder, weights_path, image_size):
+    model = builder((image_size, image_size, 3))
+    model.load_weights(weights_path)
+    return model
+
+
+def load_for_views(models, builder, num_views, image_size, weights_path):
+    if num_views not in models:
```
oarriaga: This is a strange condition that it's rather unclear about what exactly is being compared.

paz/applications/depth_estimators.py
```diff
@@ -0,0 +1,98 @@
...
+        model = builder(num_views, (image_size, image_size, 3))
+        model.load_weights(weights_path)
+        models[num_views] = model
+    return models[num_views]
```
oarriaga: The key variable name is rather strange. This should be more clear.

paz/backend/image.py
```diff
@@ -81,6 +81,39 @@ def resize(image, size, method="linear", antialias=False):
...
+    uses ``keras.ops`` so it traces inside functional models and runs eagerly.
+    Separable: interpolate rows, then columns.
+    """
+    from keras import ops
+    rows = interpolate_bilinear_axis(image, 1, size[0], ops)
+    return interpolate_bilinear_axis(rows, 2, size[1], ops)
+
+
+def interpolate_bilinear_axis(image, axis, target, ops):
+    source = image.shape[axis]
```
oarriaga, line 97: There is not need to pass "ops" if this is Keras "ops" we should just simply import it on the top and use it directly.

oarriaga, line 97: The use of variables such as source and target are too generic variable names. Let's have a more clear variable names.

paz/backend/image.py
```diff
@@ -81,6 +81,39 @@ def resize(image, size, method="linear", antialias=False):
...
+    lower, upper, weight = bilinear_sample_positions(source, target, ops)
+    low = ops.take(image, lower, axis=axis)
+    high = ops.take(image, upper, axis=axis)
+    shape = [1] * len(image.shape)
+    shape[axis] = target
+    weight = ops.reshape(weight, shape)
+    return low * (1.0 - weight) + high * weight
+
+
+def bilinear_sample_positions(source, target, ops):
```
oarriaga, line 107: No need to pass ops around. Change the names source and target to be more clear variable names that better depict the meaning.

paz/models/foundation/depth_anything3/backbone.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+def apply_local_block(tokens, num_views, positions, hidden_size, num_heads,
+                      mlp_ratio, layer_scale_init, index, alt_start):
+    folded = routing.fold_views_into_batch(tokens)
+    name = f"block_{index}"
+    if index < alt_start:
```
oarriaga: Our coding style is to use "arg" instead of "index". Moreover, "alt_start" is not full english variable name. Thus makes things unclear to a reader.

paz/models/foundation/depth_anything3/backbone.py
```diff
@@ -0,0 +1,82 @@
...
+        folded = build_dinov2_block(folded, hidden_size, num_heads, mlp_ratio,
+                                    layer_scale_init, name)
+    else:
+        folded = build_da3_block(folded, positions, hidden_size, num_heads,
+                                 mlp_ratio, layer_scale_init, name)
```
oarriaga: Maybe we can order the args that are common between this functions and partial them before or better pack them under a variable arg = (...) and then pass them unpack them inside the function as fun(*arg, new_arg, ...).

paz/models/foundation/depth_anything3/backbone.py
```diff
@@ -0,0 +1,82 @@
...
+def apply_global_block(tokens, num_views, positions, hidden_size, num_heads,
+                       mlp_ratio, layer_scale_init, index):
+    merged = routing.merge_views_into_sequence(tokens)
+    merged = build_da3_block(merged, positions, hidden_size, num_heads,
+                             mlp_ratio, layer_scale_init, f"block_{index}")
+    return routing.split_sequence_into_views(merged, num_views)
+
+
+def is_global_block(index, alt_start):
+    return index >= alt_start and index % 2 == 1
```
oarriaga: This boolean equation is rather complicated to read. The variable names should be extremely clear so that a human can read it and understand it without problems fast.

paz/models/foundation/depth_anything3/backbone.py
```diff
@@ -0,0 +1,82 @@
...
+    for feature in collected:
+        camera_tokens.append(feature[:, :, 0])
+        local_half = feature[..., :hidden_size]
+        global_half = norm(feature[..., hidden_size:])
+        normed = ops.concatenate([local_half, global_half], axis=-1)
+        features.append(normed[:, :, 1:, :])
+    return features, camera_tokens
+
+
+def build_positions(num_views, grid):
```
oarriaga: I think the name positions might be a little to generic for this function name.

paz/models/foundation/depth_anything3/backbone.py
```diff
@@ -0,0 +1,82 @@
...
+    return index >= alt_start and index % 2 == 1
+
+
+def finalize_features(collected, hidden_size):
+    norm = LayerNormalization(epsilon=LAYER_NORM_EPSILON, name="norm")
+    features, camera_tokens = [], []
+    for feature in collected:
+        camera_tokens.append(feature[:, :, 0])
+        local_half = feature[..., :hidden_size]
+        global_half = norm(feature[..., hidden_size:])
```
oarriaga: It might improve readability if you split the feature on the first line rather than doing it within  a function such as append, norm, etc.

paz/models/foundation/depth_anything3/blocks.py
```diff
@@ -0,0 +1,37 @@
+from keras.layers import Add
+
+from paz.models.transformers import attention
+from paz.models.transformers.embeddings import rotary
+from paz.models.foundation.dinov2.blocks import apply_feedforward
+from paz.models.foundation.dinov2.blocks import apply_layer_scale
+from paz.models.foundation.dinov2.blocks import normalize
+from paz.models.foundation.dinov2.blocks import project_output
+
+QK_NORM_EPSILON = 1e-5
```
oarriaga: Inline global variable name where used.

paz/models/foundation/depth_anything3/blocks.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+def build_da3_block(tokens, positions, hidden_size, num_heads, mlp_ratio,
```
oarriaga: No need to name da3 if we are already inside the depth_anythin3 model.

oarriaga: MLP should be captialized since it's an acronym.

paz/models/foundation/depth_anything3/blocks.py
```diff
@@ -0,0 +1,37 @@
...
+                    layer_scale_init, name):
```
oarriaga: The layer_scale_init might be too long of a variable name. Let's think of something else that is shorter but conveys the correct meaning and is a full English name.

paz/models/foundation/depth_anything3/blocks.py
```diff
@@ -0,0 +1,37 @@
...
+    attended = apply_da3_attention(tokens, positions, hidden_size, num_heads,
+                                   name)
+    scaled = apply_layer_scale(attended, hidden_size, layer_scale_init,
+                               f"{name}_ls1")
+    tokens = Add(name=f"{name}_add1")([tokens, scaled])
+    forwarded = apply_feedforward(tokens, hidden_size, mlp_ratio, name)
+    scaled = apply_layer_scale(forwarded, hidden_size, layer_scale_init,
+                               f"{name}_ls2")
```
oarriaga: Let's not have a split using parenthesis into multiple lines. Let's us try to have one single line always that does not break or continues below.

paz/models/foundation/depth_anything3/blocks.py
```diff
@@ -0,0 +1,37 @@
...
+
+def apply_da3_attention(tokens, positions, hidden_size, num_heads, name):
+    normed = normalize(tokens, f"{name}_norm1")
+    fused = attention.project_query_key_value(normed, hidden_size, True, name)
+    head_dim = hidden_size // num_heads
+    query, key, value = attention.split_query_key_value(fused, num_heads,
+                                                        head_dim)
+    query, key = attention.normalize_query_key(query, key, QK_NORM_EPSILON, name)
+    query = rotary.apply_2d(query, positions)
+    key = rotary.apply_2d(key, positions)
```
oarriaga: d should be capitalized.

paz/models/foundation/depth_anything3/camera.py
```diff
@@ -0,0 +1,84 @@
...
+
+Quaternions are scalar-last (xyzw). The decoder predicts a camera-to-world
+pose encoding; extrinsics are returned as the world-to-camera inverse.
+"""
+from keras import ops
+from keras.layers import Dense, ReLU
+
+from paz.models.transformers.attention import kernel
+
+FOCAL_EPSILON = 1e-6
```
oarriaga: No global variable names. Apply this for the whole PR.

paz/models/foundation/depth_anything3/camera.py
```diff
@@ -0,0 +1,84 @@
...
+def quaternion_to_matrix(quaternion):
+    i, j, k, r = ops.unstack(quaternion, axis=-1)
+    two_s = 2.0 / ops.sum(quaternion * quaternion, axis=-1)
+    row0 = ops.stack([1 - two_s * (j * j + k * k), two_s * (i * j - k * r),
+                      two_s * (i * k + j * r)], axis=-1)
+    row1 = ops.stack([two_s * (i * j + k * r), 1 - two_s * (i * i + k * k),
+                      two_s * (j * k - i * r)], axis=-1)
+    row2 = ops.stack([two_s * (i * k - j * r), two_s * (j * k + i * r),
+                      1 - two_s * (i * i + j * j)], axis=-1)
+    return ops.stack([row0, row1, row2], axis=-2)
```
oarriaga, line 52: Most likely this should exist or be reused from paz.lie.quaternion

paz/models/foundation/depth_anything3/camera.py
```diff
@@ -0,0 +1,84 @@
...
+    translation = dense(hidden, 3, "cam_dec_t")
+    quaternion = dense(hidden, 4, "cam_dec_qvec")
+    field_of_view = ReLU()(dense(hidden, 2, "cam_dec_fov"))
+    return ops.concatenate([translation, quaternion, field_of_view], axis=-1)
+
+
+def pose_encoding_to_camera(pose_encoding, image_shape):
+    translation = pose_encoding[..., :3]
+    quaternion = pose_encoding[..., 3:7]
+    field_of_view = pose_encoding[..., 7:]
```
oarriaga: Maybe we can unpack it / split this in a single line.

paz/models/foundation/depth_anything3/camera.py
```diff
@@ -0,0 +1,84 @@
...
+    transposed = ops.swapaxes(rotation, -1, -2)
+    inverse_translation = -ops.matmul(transposed, translation)
+    return ops.concatenate([transposed, inverse_translation], axis=-1)
+
+
+def intrinsics_from_field_of_view(field_of_view, image_shape):
+    height, width = image_shape[0], image_shape[1]
+    focal_y = (height / 2.0) / clamp_tan(field_of_view[..., 0])
+    focal_x = (width / 2.0) / clamp_tan(field_of_view[..., 1])
+    return stack_intrinsics(focal_x, focal_y, width / 2.0, height / 2.0)
```
oarriaga: Are there no other backend functions that perform this exact computation?

paz/models/foundation/depth_anything3/camera.py
```diff
@@ -0,0 +1,84 @@
...
+    return ops.stack([row0, row1, row2], axis=-2)
+
+
+def clamp_tan(angle):
+    return ops.maximum(ops.tan(angle / 2.0), FOCAL_EPSILON)
+
+
+def dense(tokens, units, name):
+    return Dense(units, use_bias=True, kernel_initializer=kernel(),
+                 name=name)(tokens)
```
oarriaga: Just inline the dense where is used and simply build something like dense_kwargs and pass those  directly in the inline.

paz/models/foundation/depth_anything3/dpt.py
```diff
@@ -0,0 +1,46 @@
...
+from paz.models.foundation.depth_anything3.dual_dpt import HEAD_FEATURES
+from paz.models.foundation.depth_anything3.dual_dpt import fuse_chain
+from paz.models.foundation.depth_anything3.dual_dpt import reassemble
+from paz.models.foundation.depth_anything3.dual_dpt import resize_stage
+
+
+def build_dpt(feature_maps, grid, image_shape, feature_dim, out_channels,
+              fusion_features):
+    stages = [project_stage(feature_maps[k], k, grid, feature_dim, out_channels)
+              for k in range(4)]
```
oarriaga: Make a for loop since it does not fit in a single line.

paz/models/foundation/depth_anything3/dual_dpt.py
```diff
@@ -0,0 +1,160 @@
...
+"""
+import numpy as np
+from keras import ops
+from keras.layers import Conv2D, Conv2DTranspose, LayerNormalization, ReLU
+
+from paz.backend.image import resize_bilinear_align_corners
+
+HEAD_FEATURES = 32
+POS_RATIO = 0.1
+OMEGA = 100.0
```
oarriaga: No global variables.

paz/models/foundation/depth_anything3/dual_dpt.py
```diff
@@ -0,0 +1,160 @@
...
+
+
+def build_dual_dpt(feature_maps, grid, image_shape, feature_dim, out_channels,
```
oarriaga: what is dpt? if it's an acronym it should be capitalized. Moreover since we are in the module dual_dpt maybe we should write as a comment on the top of the file and somewhere in this function what DPT stand for explicitly. Finally, since the module name and the function name we should not repeat the same name.

paz/models/foundation/depth_anything3/dual_dpt.py
```diff
@@ -0,0 +1,160 @@
...
+                   fusion_features):
+    norm = LayerNormalization(epsilon=1e-5, name="head_norm")
+    stages = [project_stage(feature_maps[k], norm, k, grid, image_shape,
+                            feature_dim, out_channels) for k in range(4)]
```
oarriaga: Make a for loop since it does not fit in a single line.

paz/models/foundation/depth_anything3/dual_dpt.py
```diff
@@ -0,0 +1,160 @@
...
+    if stage == 0:
+        return Conv2DTranspose(out_channels[0], 4, strides=4,
+                               name="head_resize_0")(features)
+    if stage == 1:
+        return Conv2DTranspose(out_channels[1], 2, strides=2,
+                               name="head_resize_1")(features)
+    if stage == 2:
+        return features
+    return Conv2D(out_channels[3], 3, strides=2, padding="same",
+                  name="head_resize_3")(features)
```
oarriaga: Functions should have a single return.

paz/models/foundation/depth_anything3/dual_dpt.py
```diff
@@ -0,0 +1,160 @@
...
+    aux = fuse_chain(lateral, "_aux", fusion_features)
+    main = Conv2D(fusion_features // 2, 3, padding="same", name="head_out1")(main)
+    aux = aux_pre_head(aux, fusion_features)
+    return main, aux
+
+
+def reassemble(features, stage, fusion_features):
+    name = f"head_layer{stage + 1}_rn"
+    return Conv2D(fusion_features, 3, padding="same", use_bias=False,
+                  name=name)(features)
```
oarriaga: Let's try to make this fit in a single line. Maybe you can build some kwargs in the line above with the name.

paz/models/foundation/depth_anything3/dual_dpt.py
```diff
@@ -0,0 +1,160 @@
...
+    hidden = Conv2D(fusion_features, 3, padding="same", name=f"{name}_conv2")(hidden)
+    return features + hidden
+
+
+def aux_pre_head(features, fusion_features):
+    half = fusion_features // 2
+    widths = (half, fusion_features, half, fusion_features, half)
+    for index, width in enumerate(widths):
+        features = Conv2D(width, 3, padding="same",
+                          name=f"head_out1_aux_{index}")(features)
```
oarriaga: Let's try to make this fit in a single line

paz/models/foundation/depth_anything3/dual_dpt.py
```diff
@@ -0,0 +1,160 @@
...
+
+
+def build_uv_grid(width, height, aspect):
+    diagonal = (aspect ** 2 + 1.0) ** 0.5
+    span_x = aspect / diagonal
+    span_y = 1.0 / diagonal
+    x = np.linspace(-span_x * (width - 1) / width,
+                    span_x * (width - 1) / width, width)
+    y = np.linspace(-span_y * (height - 1) / height,
+                    span_y * (height - 1) / height, height)
```
oarriaga: You can use W and H since they are conventions in paz.

paz/models/foundation/depth_anything3/embeddings.py
```diff
@@ -0,0 +1,39 @@
...
+    return ops.concatenate([camera, tokens[:, :, 1:, :]], axis=2)
+
+
+def broadcast_source(view_slice, token):
+    anchor = view_slice[:, :, :1, :]
+    return ops.zeros_like(anchor) + ops.expand_dims(token, axis=2)
+
+
+def build_camera_table(reference, hidden_size):
+    indices = Lambda(batched_pair_indices, output_shape=(2,),
```
oarriaga: We should not use lambda layers at all since this makes it hard to serialize the models.

paz/models/foundation/depth_anything3/embeddings.py
```diff
@@ -0,0 +1,39 @@
+from keras import ops
+from keras.layers import Embedding, Lambda
+
+from paz.models.transformers.attention import kernel
+from paz.models.foundation.dinov2.embeddings import build_dinov2_embeddings
```
oarriaga: The better idea I think is to be able to call from paz something like:
import paz
paz.dinov2.embeddings.build
That way we keep everything simple. and easy to read.

paz/models/foundation/depth_anything3/embeddings.py
```diff
@@ -0,0 +1,39 @@
...
+
+def insert_camera_tokens(tokens, hidden_size):
+    table = build_camera_table(tokens, hidden_size)
+    reference = ops.expand_dims(table[:, 0:1], axis=2)
+    source = broadcast_source(tokens[:, 1:], table[:, 1:2])
+    camera = ops.concatenate([reference, source], axis=1)
+    return ops.concatenate([camera, tokens[:, :, 1:, :]], axis=2)
+
+
+def broadcast_source(view_slice, token):
```
oarriaga: It's unclear what you mean by "source" since no variable name refers to source.

paz/models/foundation/depth_anything3/models.py
```diff
@@ -0,0 +1,116 @@
...
+from paz.models.foundation.dinov2.encoder import build_dinov2_encoder
+
+LAYER_NORM_EPSILON = 1e-6
+MONO_OUT_LAYERS = (4, 11, 17, 23)
+
+PATCH_SIZE = 14
+MLP_RATIO = 4.0
+LAYER_SCALE_INIT = 1.0
+OUT_LAYERS = (5, 7, 9, 11)
+ALT_START = 4
```
oarriaga: No global varibales. Just inline or abstract when needed / used.

paz/models/foundation/depth_anything3/models.py
```diff
@@ -0,0 +1,116 @@
...
+    return build_da3_backbone_model(*args, name)
+
+
+def build_da3_backbone_model(num_views, image_shape, hidden_size, depth,
+                             num_heads, name):
+    images = Input((num_views, *image_shape), name="views")
+    grid = grid_shape(image_shape, PATCH_SIZE)
+    num_positions = grid[0] * grid[1] + 1
+    tokens = build_view_embeddings(images, PATCH_SIZE, hidden_size,
+                                   num_positions, num_views)
```
oarriaga: We could wrap some of the inputs here into some args on the line before to unpack them here and keep all compuation on one line.

oarriaga: Same in the line below.

paz/models/foundation/depth_anything3/models.py
```diff
@@ -0,0 +1,116 @@
+from keras import Input, Model, ops
+from keras.layers import LayerNormalization
+
```
oarriaga, line 3: Most of the models here should be referred with CamelCase (all functions that return a Keras model should follow this convention). It's OK to repeat the name DepthAnything in the names. Those should be directly exported from the module so that we could call:
paz.models.DepthAnythingV3Small (maybe we need to introduce Mono, Metric, etc).

paz/models/foundation/depth_anything3/models.py
```diff
@@ -0,0 +1,116 @@
...
+    return build_da3(num_views, image_shape, *args, name)
+
+
+def build_da3(num_views, image_shape, hidden_size, num_heads, out_channels,
+              fusion_features, name):
+    images = Input((num_views, *image_shape), name="views")
+    grid = grid_shape(image_shape, PATCH_SIZE)
+    num_positions = grid[0] * grid[1] + 1
+    tokens = build_view_embeddings(images, PATCH_SIZE, hidden_size,
+                                   num_positions, num_views)
```
oarriaga: Let's add an args before so that it fits in one line everything without breaking parenthesis.

oarriaga: Same everywhere and in this function.

paz/models/foundation/depth_anything3/routing.py
```diff
@@ -0,0 +1,60 @@
...
+    patches = np.ones((num_patches, 2), "int32")
+    per_view = np.concatenate([special, patches], axis=0)
+    positions = np.tile(per_view, (num_views, 1))
+    return ops.array(positions[None])
+
+
+def build_grid_positions(grid_height, grid_width):
+    rows = np.repeat(np.arange(grid_height), grid_width)
+    columns = np.tile(np.arange(grid_width), grid_height)
+    return np.stack([rows, columns], axis=-1).astype("int32")
```
oarriaga: Why are we using numpy here directly? Shall we not use jax.numpy?

paz/models/foundation/dinov2/blocks.py
```diff
@@ -0,0 +1,54 @@
+import keras
+from keras.layers import Add, Dense, EinsumDense, LayerNormalization
+
+from paz.models.transformers import attention, feedforward
+
+LAYER_NORM_EPSILON = 1e-6
```
oarriaga: No global variables.

paz/models/foundation/dinov2/encoder.py
```diff
@@ -0,0 +1,11 @@
+from paz.models.foundation.dinov2.blocks import build_dinov2_block
+
+
+def build_dinov2_encoder(tokens, hidden_size, depth, num_heads, mlp_ratio,
```
oarriaga: No need to repeat name of the module on the function.

paz/models/foundation/dinov2/encoder.py
```diff
@@ -0,0 +1,11 @@
...
+                         layer_scale_init):
+    hidden_states = []
+    for index in range(depth):
```
oarriaga: We should use arg instead of index.

paz/models/foundation/dinov2/encoder.py
```diff
@@ -0,0 +1,11 @@
...
+        args = (hidden_size, num_heads, mlp_ratio, layer_scale_init)
+        tokens = build_dinov2_block(tokens, *args, f"block_{index}")
```
oarriaga: No need again to use build_dinov2_block. Ideally we would have something like:
paz.dinov2.blocks.build

paz/models/transformers/attention.py
```diff
@@ -172,3 +172,37 @@ def merge_heads(tensor):
 
 def kernel(stddev=0.02):
     return keras.initializers.TruncatedNormal(stddev=stddev)
+
+
+def project_query_key_value(tokens, hidden_size, use_bias, name):
+    units = hidden_size * 3
+    layer = Dense(units, use_bias=use_bias, kernel_initializer=kernel(),
+                  name=f"{name}_qkv")
```
oarriaga: Use kwargs in the line before to not break the line in the Dense

paz/models/transformers/embeddings/rotary.py
```diff
@@ -13,6 +13,29 @@ def apply(inputs, wavelength, scaling_factor, denominator, positions=None):
     return (inputs * cosine) + (rotated * sine)
 
 
+def apply_2d(tokens, positions, base_frequency=100.0):
```
oarriaga: apply2D

paz/models/transformers/embeddings/rotary.py
```diff
@@ -13,6 +13,29 @@ def apply(inputs, wavelength, scaling_factor, denominator, positions=None):
...
+    vertical = apply_axis(vertical, positions[..., 0], base_frequency)
+    horizontal = apply_axis(horizontal, positions[..., 1], base_frequency)
+    return ops.concatenate([vertical, horizontal], axis=-1)
+
+
+def apply_axis(tokens, positions, base_frequency):
+    half = tokens.shape[-1]
+    exponents = ops.arange(0, half, 2, dtype="float32") / half
+    inverse = 1.0 / (base_frequency ** exponents)
+    angles = ops.cast(positions, "float32")[..., None] * inverse
```
oarriaga, line 27: Maybe we should use expand_dims directly and split it up into multiple lines.

paz/models/foundation/dinov2/models.py
```diff
@@ -0,0 +1,102 @@
...
+
+from paz.models.foundation.dinov2.embeddings import build_dinov2_embeddings
+from paz.models.foundation.dinov2.encoder import build_dinov2_encoder
+
+PATCH_SIZE = 14
+MLP_RATIO = 4.0
+LAYER_SCALE_INIT = 1e-5
+LAYER_NORM_EPSILON = 1e-6
+DEFAULT_IMAGE_SHAPE = (518, 518, 3)
+DEFAULT_OUT_LAYERS = (5, 7, 9, 11)
```
oarriaga: No global variables.

oarriaga: Inline them where is needed or better yet abstract them.

paz/models/foundation/dinov2/models.py
```diff
@@ -0,0 +1,102 @@
...
+
+
+def build_dinov2(image_shape, patch_size, hidden_size, depth, num_heads,
+                 mlp_ratio, num_register_tokens, name):
+    images = Input(image_shape, name="pixels")
+    positions = count_positions(image_shape, patch_size)
+    tokens = build_dinov2_embeddings(images, patch_size, hidden_size,
+                                     positions, num_register_tokens)
```
oarriaga: Let's avoid breaking maybe just have a line before that packs things into a list as args and then unpack in the second line.

oarriaga: Apply the same in this file where we have a similar issue.

paz/models/foundation/dinov2/embeddings.py
```diff
@@ -0,0 +1,50 @@
+from keras import ops
+from keras.layers import Conv2D, Embedding, Lambda, Reshape
+
+from paz.models.transformers.attention import kernel
+from paz.models.transformers.embeddings import absolute
+
+
+def build_dinov2_embeddings(images, patch_size, hidden_size, num_positions,
```
oarriaga: We should change this function name such that we could something like:
paz.dinov2.embeddings.build

paz/models/foundation/dinov2/embeddings.py
```diff
@@ -0,0 +1,50 @@
...
+    tokens = add_position_embedding(tokens, num_positions)
+    if num_register_tokens:
+        tokens = insert_register_tokens(tokens, num_register_tokens,
+                                        hidden_size)
+    return tokens
+
+
+def build_patch_tokens(images, patch_size, hidden_size):
+    projection = Conv2D(hidden_size, patch_size, strides=patch_size,
+                        padding="valid", name="patch_embed_proj")
```
oarriaga: Let's build some kwargs in the line before to avoid using parenthesis across multiple lines.

oarriaga:
Addressed the review in `bb7b1140`. Grouped resolution:

- **No module-level globals** (items 1,12,29,35,41,51,58,65): inlined literals or moved them into the size-specific builders.
- **80 cols, no paren-continuations** (6,7,11,33,40,43,45,46,52,55,62,67,70): packed inputs into `args`/`kwargs` on the preceding line and used real for-loops. Enforced by the repo line-length hook.
- **paz.image, not OpenCV** (2,8,9,10): applications and examples now use `paz.image.load` / `resize_opencv` / `standardize` / `write`.
- **backend/image resize** (3,4,5,21,22,23): `keras.ops` imported at module top, no passed-`ops`, clearer names, single-purpose helpers.
- **Naming** (5,14,16-20,22,24,27,30-34,42,47,50,59,60,63): `arg` over `index`, acronyms capitalized (`MLP_ratio`, `apply_2D`), dropped module-name prefixes so functions read as `module.build`, `W`/`H`, unpacked shapes once, clearer variable names.
- **Split large functions / single return** (13,15,25,26,28,44): e.g. the backbone encoder and DualDPT.
- **CamelCase + export** (54): `DepthAnything3Small/Base/MonoLarge/MetricLarge`, exported from `paz.models`; also added `paz.dinov2.embeddings.build`-style access (49,61,69).
- **jax.numpy** (57): `routing` positions now use `jax.numpy`.

Three items involved a judgment call:

1. **Lambda layers (48):** removed everywhere. The only classless way to hold a learnable token in Keras still needs a Lambda, so I introduced one small serializable `LearnableTokens` layer (`transformers/embeddings/token.py`, `add_weight` + `get_config`), shared by DINOv2 and DA3 — the same pattern as dinov3/gemma4's `PositionEmbeddingTable`. This relaxes the "functions only" rule for this one primitive to gain serializability, which was the point of the comment.
2. **Camera reuse (36,38):** `paz.angles.quaternion_to_rotation_matrix` (xyzw), `paz.SE3.invert`, and `paz.pinhole` do exist, but they are unbatched jax/numpy and can't run inside the batched Keras functional graph where DA3 predicts cameras. So `camera.py` keeps small, explicit, tested `keras.ops` versions (a comment notes this). Happy to move the geometry to the application layer instead if you'd prefer maximizing reuse.
3. **Naming (54):** used `DepthAnything3*` (matching the `depth_anything3` module and the official "Depth Anything 3") rather than the literal `DepthAnythingV3*` — let me know if you want `V3`.

Verification: `dinov2 == legacy` bit-for-bit unchanged; DA3 SMALL/BASE/MONO still match the reference (opt-in `DA3_REFERENCE`). Suite 75 passed / 5 skipped; 80-column and pyflakes clean.

PR #405, opened 2026-07-15
Port SONIC deploy actor from Keras to paz style

oarriaga:
## Why

The SONIC humanoid whole-body-control actor (three encoder branches with an
FSQ-quantizer bottleneck, feeding a decoder MLP) was implemented as a Keras
port in `gear_sonic/keras` (GR00T-WholeBodyControl), independent of paz's
conventions. This brings that architecture and its ONNX weight-import
tooling into paz as `paz/models/foundation/sonic`.

## What changed

- `layout.py`: `SonicObservationLayout` / `EncoderModeLayout` /
  `ObservationSpan` as namedtuples with `compute_*`/`find_*` helper
  functions, replacing the original's dataclass properties/methods.
  `load_release_observation_layout` parses the release YAML unchanged.
- `model.py`: `build_sonic_encoder` / `build_sonic_decoder` /
  `build_sonic_actor` plus the FSQ/dense-stack/mode-routing helpers,
  restyled but behaviorally identical to the original. FSQ hyperparameters
  are now a named `FSQArgs` tuple instead of an anonymous 6-tuple.
- `conversion.py`: imports the released ONNX weights into the ported Keras
  models via the same regex-based layer-name matching as the original
  (encoder kernels transposed, decoder kernels not - that asymmetry is
  intentional and preserved). Includes a CLI (`main()`) that writes a
  complete asset bundle (both weight files plus `observation_config.yaml`).
- `pretrained.py`: `Sonic(weights="pretrained")` follows the same
  `keras.utils.get_file` download pattern as `Gemma4`/`Whisper`. The
  converted weights are not uploaded to `altamira-data` yet - see the
  module docstring; publishing them is a follow-up decision, not part of
  this PR.
- Registered `Sonic` in `paz/models/__init__.py`.

## Verification

- `pytest paz/models/foundation/sonic/` -> 17 passed, 2 skipped.
- The 2 skipped tests are an opt-in numeric-parity check (gated by
  `SONIC_RELEASE_DIR`) against the real released ONNX files. Run locally
  against those assets: encoder output matches ONNX Runtime exactly (max
  abs diff 0.0) across all three modes (g1/teleop/smpl); decoder matches
  within 1e-5, consistent with the tolerances already validated in the
  source repo.
- `pyflakes` clean on all new files.
- `/code-review` at medium effort found and fixed 4 issues: the conversion
  CLI's output bundle was missing `observation_config.yaml`, a
  hanging-parenthesis FSQ signature/anonymous tuple, duplicated test
  fixtures across two test files, and a `conftest.py` convention drift
  from the sibling `gemma4` package.

### DIFF
--- a/paz/models/foundation/sonic/conversion.py
+++ b/paz/models/foundation/sonic/conversion.py
@@ -1,0 +1,122 @@
+"""Import released SONIC ONNX weights into the ported Keras models."""
+
+from argparse import ArgumentParser
+from pathlib import Path
+import re
+import shutil
+
+import onnx
+from onnx import numpy_helper
+
+from paz.models.foundation.sonic.layout import compute_decoder_input_dim
+from paz.models.foundation.sonic.layout import compute_encoder_input_dim
+from paz.models.foundation.sonic.layout import load_release_observation_layout
+from paz.models.foundation.sonic.model import build_sonic_actor
+from paz.models.foundation.sonic.model import build_sonic_decoder
+from paz.models.foundation.sonic.model import build_sonic_encoder
+
+_ENCODER_BRANCH_PATTERN = re.compile(
+    r"^module\.encoders\.(?P<branch>[^.]+)\.module\.(?P<layer>\d+)"
+    r"\.(?P<kind>weight|bias)$")
+_DECODER_NODE_PATTERN = re.compile(
+    r"^/g1_dyn/module/module\.(?P<layer>\d+)/MatMul$")
+
+
+def port_sonic_weights(layout, encoder_onnx_path, decoder_onnx_path):
+    encoder = build_sonic_encoder(layout)
+    decoder = build_sonic_decoder(layout)
+    check_input_dim(encoder, compute_encoder_input_dim(layout), "encoder")
+    check_input_dim(decoder, compute_decoder_input_dim(layout), "decoder")
+    apply_encoder_weights(encoder, load_onnx_model(encoder_onnx_path))
+    apply_decoder_weights(decoder, load_onnx_model(decoder_onnx_path))
+    actor = build_sonic_actor(layout, encoder, decoder)
+    return encoder, decoder, actor
+
+
+def check_input_dim(model, expected_dim, name):
+    input_dim = model.input_shape[-1]
+    if input_dim != expected_dim:
+        raise ValueError(
+            f"{name} input mismatch: expected {expected_dim}, got "
+            f"{input_dim}")
+
+
+def load_onnx_model(model_path):
+    return onnx.load(str(model_path))
+
+
+def load_onnx_initializers(model):
+    return {tensor.name: numpy_helper.to_array(tensor)
+            for tensor in model.graph.initializer}
+
+
+def apply_encoder_weights(encoder, encoder_onnx):
+    initializers = load_onnx_initializers(encoder_onnx)
+    for name, value in initializers.items():
+        match = _ENCODER_BRANCH_PATTERN.match(name)
+        if match is None:
+            continue
+        branch, layer = match.group("branch"), match.group("layer")
+        layer_name = f"{branch}_module_{layer}"
+        set_dense_weight(encoder.get_layer(layer_name), match.group("kind"),
+                          value)
+
+
+def set_dense_weight(layer, kind, value):
+    kernel, bias = layer.get_weights()
+    if kind == "weight":
+        kernel = value.transpose(1, 0)
+    else:
+        bias = value
+    layer.set_weights([kernel, bias])
+
+
+def apply_decoder_weights(decoder, decoder_onnx):
+    initializers = load_onnx_initializers(decoder_onnx)
+    kernel_names = find_decoder_kernel_names(decoder_onnx)
+    for layer_id, kernel_name in kernel_names.items():
+        layer = decoder.get_layer(f"g1_dyn_module_{layer_id}")
+        bias_name = f"module.decoders.g1_dyn.module.{layer_id}.bias"
+        layer.set_weights(
+            [initializers[kernel_name], initializers[bias_name]])
+
+
+def find_decoder_kernel_names(decoder_onnx):
+    kernel_names = {}
+    for node in decoder_onnx.graph.node:
+        match = _DECODER_NODE_PATTERN.match(node.name)
+        if match is not None:
+            kernel_names[int(match.group("layer"))] = node.input[1]
+    return kernel_names
+
+
+def save_ported_weights(encoder, decoder, obs_config_path, output_dir):
+    output_dir = Path(output_dir)
+    output_dir.mkdir(parents=True, exist_ok=True)
+    encoder.save_weights(output_dir / "sonic_encoder.weights.h5")
+    decoder.save_weights(output_dir / "sonic_decoder.weights.h5")
+    shutil.copy(obs_config_path, output_dir / "observation_config.yaml")
+    return output_dir
+
+
+def build_argument_parser():
+    parser = ArgumentParser()
+    parser.add_argument("--obs_config", required=True)
+    parser.add_argument("--encoder_onnx", required=True)
+    parser.add_argument("--decoder_onnx", required=True)
+    parser.add_argument("--output_dir", default="~/.keras/paz/models/sonic")
+    return parser
+
+
+def main():
+    args = build_argument_parser().parse_args()
+    layout = load_release_observation_layout(args.obs_config)
+    encoder, decoder, _ = port_sonic_weights(
+        layout, args.encoder_onnx, args.decoder_onnx)
+    output_dir = save_ported_weights(
+        encoder, decoder, args.obs_config, Path(args.output_dir).expanduser())
+    print(f"Saved ported SONIC weights to {output_dir}")
+
+
+if __name__ == "__main__":
+    main()
--- a/paz/models/foundation/sonic/conversion.py
+++ b/paz/models/foundation/sonic/conversion.py
@@ -35,9 +35,7 @@ def port_weights(layout, encoder_onnx_path, decoder_onnx_path):
 def check_input_dim(model, expected_dim, name):
     input_dim = model.input_shape[-1]
     if input_dim != expected_dim:
-        raise ValueError(
-            f"{name} input mismatch: expected {expected_dim}, got "
-            f"{input_dim}")
+        raise ValueError(f"{name} dim mismatch: {input_dim} != {expected_dim}")
 
 
 def load_onnx_model(model_path):
@@ -45,8 +43,10 @@ def load_onnx_model(model_path):
 
 
 def load_onnx_initializers(model):
-    return {tensor.name: numpy_helper.to_array(tensor)
-            for tensor in model.graph.initializer}
+    initializers = {}
+    for tensor in model.graph.initializer:
+        initializers[tensor.name] = numpy_helper.to_array(tensor)
+    return initializers
 
 
 def apply_encoder_weights(encoder, encoder_onnx):
@@ -55,10 +55,11 @@ def apply_encoder_weights(encoder, encoder_onnx):
         match = _ENCODER_BRANCH_PATTERN.match(name)
         if match is None:
             continue
-        branch, layer = match.group("branch"), match.group("layer")
-        layer_name = f"{branch}_module_{layer}"
-        set_dense_weight(encoder.get_layer(layer_name), match.group("kind"),
-                          value)
+        branch, layer_id = match.group("branch"), match.group("layer")
+        layer_name = f"{branch}_module_{layer_id}"
+        layer = encoder.get_layer(layer_name)
+        kind = match.group("kind")
+        set_dense_weight(layer, kind, value)
 
 
 def set_dense_weight(layer, kind, value):
@@ -76,8 +77,8 @@ def apply_decoder_weights(decoder, decoder_onnx):
     for layer_id, kernel_name in kernel_names.items():
         layer = decoder.get_layer(f"g1_dyn_module_{layer_id}")
         bias_name = f"module.decoders.g1_dyn.module.{layer_id}.bias"
-        layer.set_weights(
-            [initializers[kernel_name], initializers[bias_name]])
+        weights = [initializers[kernel_name], initializers[bias_name]]
+        layer.set_weights(weights)
 
 
 def find_decoder_kernel_names(decoder_onnx):
@@ -109,10 +110,10 @@ def build_argument_parser():
 def main():
     args = build_argument_parser().parse_args()
     layout = load_release_observation_layout(args.obs_config)
-    encoder, decoder, _ = port_weights(
-        layout, args.encoder_onnx, args.decoder_onnx)
-    output_dir = save_ported_weights(
-        encoder, decoder, Path(args.output_dir).expanduser())
+    encoder_onnx, decoder_onnx = args.encoder_onnx, args.decoder_onnx
+    encoder, decoder, _ = port_weights(layout, encoder_onnx, decoder_onnx)
+    output_dir = Path(args.output_dir).expanduser()
+    output_dir = save_ported_weights(encoder, decoder, output_dir)
     print(f"Saved ported SONIC weights to {output_dir}")
 
 
--- a/paz/models/foundation/sonic/conversion.py
+++ b/paz/models/foundation/sonic/conversion.py
@@ -11,9 +11,9 @@ from onnx import numpy_helper
 from paz.models.foundation.sonic.layout import compute_decoder_input_dim
 from paz.models.foundation.sonic.layout import compute_encoder_input_dim
 from paz.models.foundation.sonic.layout import load_release_observation_layout
-from paz.models.foundation.sonic.model import build_sonic_actor
-from paz.models.foundation.sonic.model import build_sonic_decoder
-from paz.models.foundation.sonic.model import build_sonic_encoder
+from paz.models.foundation.sonic.model import build_actor
+from paz.models.foundation.sonic.model import build_decoder
+from paz.models.foundation.sonic.model import build_encoder
 
 _ENCODER_BRANCH_PATTERN = re.compile(
     r"^module\.encoders\.(?P<branch>[^.]+)\.module\.(?P<layer>\d+)"
@@ -22,14 +22,14 @@ _DECODER_NODE_PATTERN = re.compile(
     r"^/g1_dyn/module/module\.(?P<layer>\d+)/MatMul$")
 
 
-def port_sonic_weights(layout, encoder_onnx_path, decoder_onnx_path):
-    encoder = build_sonic_encoder(layout)
-    decoder = build_sonic_decoder(layout)
+def port_weights(layout, encoder_onnx_path, decoder_onnx_path):
+    encoder = build_encoder(layout)
+    decoder = build_decoder(layout)
     check_input_dim(encoder, compute_encoder_input_dim(layout), "encoder")
     check_input_dim(decoder, compute_decoder_input_dim(layout), "decoder")
     apply_encoder_weights(encoder, load_onnx_model(encoder_onnx_path))
     apply_decoder_weights(decoder, load_onnx_model(decoder_onnx_path))
-    actor = build_sonic_actor(layout, encoder, decoder)
+    actor = build_actor(layout, encoder, decoder)
     return encoder, decoder, actor
 
 
@@ -111,7 +111,7 @@ def build_argument_parser():
 def main():
     args = build_argument_parser().parse_args()
     layout = load_release_observation_layout(args.obs_config)
-    encoder, decoder, _ = port_sonic_weights(
+    encoder, decoder, _ = port_weights(
         layout, args.encoder_onnx, args.decoder_onnx)
     output_dir = save_ported_weights(
         encoder, decoder, args.obs_config, Path(args.output_dir).expanduser())
--- a/paz/models/foundation/sonic/conversion.py
+++ b/paz/models/foundation/sonic/conversion.py
@@ -3,7 +3,6 @@
 from argparse import ArgumentParser
 from pathlib import Path
 import re
-import shutil
 
 import onnx
 from onnx import numpy_helper
@@ -90,12 +89,11 @@ def find_decoder_kernel_names(decoder_onnx):
     return kernel_names
 
 
-def save_ported_weights(encoder, decoder, obs_config_path, output_dir):
+def save_ported_weights(encoder, decoder, output_dir):
     output_dir = Path(output_dir)
     output_dir.mkdir(parents=True, exist_ok=True)
     encoder.save_weights(output_dir / "sonic_encoder.weights.h5")
     decoder.save_weights(output_dir / "sonic_decoder.weights.h5")
-    shutil.copy(obs_config_path, output_dir / "observation_config.yaml")
     return output_dir
 
 
@@ -114,7 +112,7 @@ def main():
     encoder, decoder, _ = port_weights(
         layout, args.encoder_onnx, args.decoder_onnx)
     output_dir = save_ported_weights(
-        encoder, decoder, args.obs_config, Path(args.output_dir).expanduser())
+        encoder, decoder, Path(args.output_dir).expanduser())
     print(f"Saved ported SONIC weights to {output_dir}")
 
 
--- a/paz/models/foundation/sonic/layout.py
+++ b/paz/models/foundation/sonic/layout.py
@@ -115,20 +115,20 @@ def load_release_observation_layout(obs_config_path):
     check_policy_spans(policy_spans, token_dim)
     check_encoder_spans(encoder_spans)
     encoder_span_map = {span.name: span for span in encoder_spans}
-    mode_configs = encoder_config.get("encoder_modes", [])
-    encoder_modes = [
-        load_encoder_mode(mode_config, encoder_span_map)
-        for mode_config in mode_configs
-    ]
+    encoder_modes = load_encoder_modes(encoder_config, encoder_span_map)
+    return SonicObservationLayout(
+        policy_spans, encoder_spans, encoder_modes, token_dim)
+
+
+def load_encoder_modes(encoder_config, encoder_span_map):
+    encoder_modes = []
+    for mode_config in encoder_config.get("encoder_modes", []):
+        mode = load_encoder_mode(mode_config, encoder_span_map)
+        encoder_modes.append(mode)
     if not encoder_modes:
         raise ValueError("No encoder_modes found in observation config")
-    args = (
-        policy_spans,
-        encoder_spans,
-        tuple(sorted(encoder_modes, key=lambda mode: mode.mode_id)),
-        token_dim,
-    )
-    return SonicObservationLayout(*args)
+    sorted_modes = sorted(encoder_modes, key=lambda mode: mode.mode_id)
+    return tuple(sorted_modes)
 
 
 def load_yaml(obs_config_path):
@@ -145,8 +145,7 @@ def check_policy_spans(policy_spans, token_dim):
         raise ValueError(f"Expected 'token_state' first, got {name}")
     if policy_spans[0].dim != token_dim:
         dim = policy_spans[0].dim
-        raise ValueError(
-            f"token_state dimension mismatch: expected {token_dim}, got {dim}")
+        raise ValueError(f"token_state dim mismatch: {dim} != {token_dim}")
 
 
 def check_encoder_spans(encoder_spans):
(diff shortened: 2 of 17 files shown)


paz/models/foundation/sonic/layout.py
```diff
@@ -0,0 +1,228 @@
...
+        for mode_config in mode_configs
+    ]
+    if not encoder_modes:
+        raise ValueError("No encoder_modes found in observation config")
+    args = (
+        policy_spans,
+        encoder_spans,
+        tuple(sorted(encoder_modes, key=lambda mode: mode.mode_id)),
+        token_dim,
+    )
```
oarriaga: This should be a cleaned to always fit on single lines. Maybe split the for loop. Split the tuple, sorted computation etc.

paz/models/foundation/sonic/layout.py
```diff
@@ -0,0 +1,228 @@
...
+def check_policy_spans(policy_spans, token_dim):
+    if not policy_spans:
+        raise ValueError("No enabled policy observations found")
+    if policy_spans[0].name != "token_state":
+        name = policy_spans[0].name
+        raise ValueError(f"Expected 'token_state' first, got {name}")
+    if policy_spans[0].dim != token_dim:
+        dim = policy_spans[0].dim
+        raise ValueError(
+            f"token_state dimension mismatch: expected {token_dim}, got {dim}")
```
oarriaga: No line break.

paz/models/foundation/sonic/layout.py
```diff
@@ -0,0 +1,228 @@
...
+        raise ValueError(f"Expected 'encoder_mode_4' first, got {name}")
+
+
+def load_encoder_mode(mode_config, encoder_span_map):
+    required_observations = tuple(mode_config.get("required_observations", []))
+    feature_spans = tuple(
+        encoder_span_map[name]
+        for name in required_observations
+        if name != "encoder_mode_4"
+    )
```
oarriaga: Split into a normal for loop with if condition

paz/models/foundation/sonic/layout.py
```diff
@@ -0,0 +1,228 @@
...
+    feature_groups = compute_feature_groups(
+        mode_config["name"], feature_spans, encoder_span_map)
+    args = (
+        mode_config["name"],
+        int(mode_config["mode_id"]),
+        required_observations,
+        feature_spans,
+        temporal_frames,
+        feature_groups,
+    )
```
oarriaga: Don't break computation over lines.

paz/models/foundation/sonic/layout.py
```diff
@@ -0,0 +1,228 @@
...
+    return None
+
+
+def compute_feature_groups(mode_name, feature_spans, encoder_span_map):
+    if mode_name not in _RELEASE_ENCODER_GROUPS:
+        return tuple((span,) for span in feature_spans)
+    return tuple(
+        tuple(encoder_span_map[name] for name in group)
+        for group in _RELEASE_ENCODER_GROUPS[mode_name]
+    )
```
oarriaga: Split into a single for loop.

paz/models/foundation/sonic/layout.py
```diff
@@ -0,0 +1,228 @@
...
+        raise ValueError(f"Unsupported observation name: {name}")
+    base_name = match.group("base")
+    if base_name not in _BASE_OBSERVATION_DIMS:
+        raise ValueError(f"Unsupported base observation name: {base_name}")
+    return _BASE_OBSERVATION_DIMS[base_name] * int(match.group("frames"))
+
+
+def load_enabled_spans(observation_entries, token_dim):
+    spans = []
+    offset = 0
```
oarriaga: You can put this two in a single line.

paz/models/foundation/sonic/model.py
```diff
@@ -0,0 +1,191 @@
...
+from paz.models.foundation.sonic.layout import EncoderModeLayout
+from paz.models.foundation.sonic.layout import ObservationSpan
+from paz.models.foundation.sonic.layout import SonicObservationLayout
+from paz.models.foundation.sonic.layout import compute_decoder_input_dim
+from paz.models.foundation.sonic.layout import compute_encoder_input_dim
+from paz.models.foundation.sonic.layout import compute_mode_scalar_index
+from paz.models.foundation.sonic.layout import compute_policy_tail_dim
+
+_ENCODER_HIDDEN_DIMS = (2048, 1024, 512, 512)
+_DECODER_HIDDEN_DIMS = (2048, 2048, 1024, 1024, 512, 512)
```
oarriaga: No global variables. Inline where needed.

paz/models/foundation/sonic/model.py
```diff
@@ -0,0 +1,191 @@
...
+
+FSQArgs = namedtuple(
+    "FSQArgs", "num_tokens token_dim offset scale rounding_shift divisor")
+_RELEASE_FSQ_ARGS = FSQArgs(2, 32, 0.032237, 15.515501, 0.5, 16.0)
+
+
+def build_sonic_actor(layout, encoder, decoder):
```
oarriaga: It's already clear that we are already under the sonic moduel foundation.sonic.model so there is no need to name it "sonic" in the function def. This applies everywhere in this PR as well.

paz/models/foundation/sonic/model.py
```diff
@@ -0,0 +1,191 @@
...
+    encoder_obs = build_input("encoder_obs", compute_encoder_input_dim(layout))
+    policy_obs = build_input(
+        "policy_obs_tail", compute_policy_tail_dim(layout))
```
oarriaga: split into two lines so that everything fits in less than 80 columns

paz/models/foundation/sonic/model.py
```diff
@@ -0,0 +1,191 @@
...
+    inputs = {"encoder_obs": encoder_obs, "policy_obs_tail": policy_obs}
+    return keras.Model(inputs, action, name="sonic_actor")
+
+
+def build_sonic_encoder(layout):
+    encoder_input = build_input("obs_dict", compute_encoder_input_dim(layout))
+    mode_id = compute_mode_id(encoder_input, compute_mode_scalar_index(layout))
+    branch_tokens = compute_branch_tokens_list(encoder_input, layout)
+    selected_tokens = compute_selected_tokens(
+        branch_tokens, mode_id, len(layout.encoder_modes))
```
oarriaga: split into two lines so that everything fits in less than 80 columns

paz/models/foundation/sonic/model.py
```diff
@@ -0,0 +1,191 @@
...
+    return keras.Model(encoder_input, selected_tokens, name="sonic_encoder")
+
+
+def build_sonic_decoder(layout):
+    decoder_input = build_input("obs_dict", compute_decoder_input_dim(layout))
+    decoder_features = compute_decoder_features(
+        decoder_input, layout.token_dim, compute_policy_tail_dim(layout))
```
oarriaga: split into two lines so that everything fits in less than 80 columns

paz/models/foundation/sonic/model.py
```diff
@@ -0,0 +1,191 @@
...
+def build_sonic_encoder(layout):
+    encoder_input = build_input("obs_dict", compute_encoder_input_dim(layout))
+    mode_id = compute_mode_id(encoder_input, compute_mode_scalar_index(layout))
+    branch_tokens = compute_branch_tokens_list(encoder_input, layout)
+    selected_tokens = compute_selected_tokens(
+        branch_tokens, mode_id, len(layout.encoder_modes))
+    return keras.Model(encoder_input, selected_tokens, name="sonic_encoder")
+
+
+def build_sonic_decoder(layout):
```
oarriaga: It's already clear that we are already under the sonic module foundation.sonic.model so there is no need to name it "sonic" in the function def. This applies everywhere in this PR as well.

paz/models/foundation/sonic/model.py
```diff
@@ -0,0 +1,191 @@
...
+
+def compute_decoder_features(decoder_input, token_dim, tail_dim):
+    token_state = decoder_input[:, :token_dim]
+    policy_tail = decoder_input[:, -tail_dim:]
+    return compute_cat(token_state, policy_tail)
+
+
+def compute_dense_stack(inputs, hidden_dims, output_dim, prefix):
+    x = inputs
+    layer_id = 0
```
oarriaga: fits in a single line.

paz/models/foundation/sonic/model.py
```diff
@@ -0,0 +1,191 @@
...
+    for units in hidden_dims:
+        x = layers.Dense(units, name=f"{prefix}_module_{layer_id}")(x)
+        x = layers.Activation("swish", name=f"{prefix}_silu_{layer_id}")(x)
+        layer_id += 2
```
oarriaga: we are explicit about layer_id = layer_id + 2 moreover we don't use id if we could use arg.

paz/models/foundation/sonic/model.py
```diff
@@ -0,0 +1,191 @@
...
+
+
+def compute_cat(left, right):
+    return keras.ops.concatenate([left, right], axis=-1)
+
+
+def compute_release_fsq(inputs, fsq_args):
+    inputs = keras.ops.cast(inputs, "float32")
+    inputs = keras.ops.reshape(
+        inputs, (-1, fsq_args.num_tokens, fsq_args.token_dim))
```
oarriaga: Split computation. Maybe define the shape before.

paz/models/foundation/sonic/pretrained.py
```diff
@@ -0,0 +1,47 @@
...
+from paz.models.foundation.sonic.model import build_sonic_decoder
+from paz.models.foundation.sonic.model import build_sonic_encoder
+
+SONIC_ASSETS_URL = "https://github.com/oarriaga/altamira-data/releases/download/v0.27/"  # fmt: skip
+SONIC_CACHE_SUBDIR = "paz/models/sonic"
+
+SonicModels = namedtuple("Sonic", "layout encoder decoder actor")
+
+
+def Sonic(weights="pretrained"):
```
oarriaga: I think the original model is capitalized SONIC we should enforce this in the  API.

paz/models/foundation/sonic/model.py
```diff
@@ -0,0 +1,191 @@
...
+    inputs = keras.ops.tanh(inputs + fsq_args.offset) * fsq_args.scale
+    inputs = keras.ops.round(inputs - fsq_args.rounding_shift)
+    inputs = inputs / fsq_args.divisor
+    return keras.ops.reshape(
+        inputs, (-1, fsq_args.num_tokens * fsq_args.token_dim))
+
+
+def build_toy_layout():
```
oarriaga: Is this toy layout needed?

oarriaga:
Addressed in 95ec3e48 and 6b822e9a:

- **Redundant "sonic" in names** (applies everywhere): `build_sonic_actor/encoder/decoder` -> `build_actor/encoder/decoder`, `port_sonic_weights` -> `port_weights`, `fetch_sonic_asset` -> `fetch_asset`. Left `SonicObservationLayout`/`SonicModels` as-is (matches the `Gemma4Backbone`/`Gemma4Models` precedent of a model's identity types keeping its name).
- **`Sonic()` -> `SONIC()`**, matching the original model's acronym casing.
- **No global variables**: inlined `_ENCODER_HIDDEN_DIMS`/`_DECODER_HIDDEN_DIMS`/`_RELEASE_FSQ_ARGS` at their single call sites instead of module scope.
- **`layer_id += 2`**: renamed to `layer_index` with an explicit `layer_index = layer_index + 2`.
- **layout.py wrapping**: replaced the wrapped tuples/comprehensions in `load_release_observation_layout`, `load_encoder_mode`, and `compute_feature_groups` with explicit for loops / direct constructor calls; shortened the token_state mismatch message so it fits on one line without breaking; combined `spans, offset = [], 0`.
- **`compute_release_fsq` "split computation"**: named the reshape target shapes (`token_shape`, `flat_shape`) instead of wrapping the `reshape` calls.
- **"Is this toy layout needed?"**: moved `build_toy_layout` out of `model.py` (production code) entirely, into a `toy_layout` pytest fixture in `conftest.py`, shared by `model_test.py`/`conversion_test.py`. So no, it wasn't needed in the model module — good catch.
- The three "split into two lines" comments on the `build_input(...)`/`compute_selected_tokens(...)`/`compute_decoder_features(...)` calls: these were already split across two lines and within 80 columns (verified: combining them onto one line would be 80, 96, and 113 characters respectively), so no further change there.

Also proposed 3 new CLAUDE.md/AGENTS.md rules from the generalizable feedback (module-name repetition, acronym casing, single-use module constants) - held those for your direct review rather than auto-applying, and per your call they're staying as local, uncommitted edits since CLAUDE.md/AGENTS.md aren't tracked in this repo right now.

Full suite: `pytest paz/models/foundation/sonic/ examples/sonic/` -> 28 passed, 2 skipped (opt-in real-weight parity tests, which pass locally against the real release assets).

paz/models/foundation/sonic/model.py
```diff
@@ -0,0 +1,162 @@
...
+    decoder_features = compute_decoder_features(
+        decoder_input, layout.token_dim, compute_policy_tail_dim(layout))
+    hidden_dims = (2048, 2048, 1024, 1024, 512, 512)
+    args = (decoder_features, hidden_dims, layout.action_dim, "g1_dyn")
+    action = compute_dense_stack(*args)
+    return keras.Model(decoder_input, action, name="sonic_decoder")
+
+
+def build_input(name, dim):
+    return keras.Input(shape=(dim,), dtype="float32", name=name)
```
oarriaga: Just inline this where is being used. No need to abstract away here since it's clear what the Input layer is doing.

paz/models/foundation/sonic/model.py
```diff
@@ -0,0 +1,162 @@
...
+from paz.models.foundation.sonic.layout import compute_policy_tail_dim
+
+FSQArgs = namedtuple(
+    "FSQArgs", "num_tokens token_dim offset scale rounding_shift divisor")
+
+
+def build_actor(layout, encoder, decoder):
+    encoder_obs = build_input("encoder_obs", compute_encoder_input_dim(layout))
+    policy_obs = build_input(
+        "policy_obs_tail", compute_policy_tail_dim(layout))
```
oarriaga: Split into two lines to make everything fit in a single line.

paz/models/foundation/sonic/model.py
```diff
@@ -0,0 +1,162 @@
...
+import keras
+from keras import layers
+
+from paz.models.foundation.sonic.layout import compute_decoder_input_dim
+from paz.models.foundation.sonic.layout import compute_encoder_input_dim
+from paz.models.foundation.sonic.layout import compute_mode_scalar_index
+from paz.models.foundation.sonic.layout import compute_policy_tail_dim
+
+FSQArgs = namedtuple(
+    "FSQArgs", "num_tokens token_dim offset scale rounding_shift divisor")
```
oarriaga: Do we really need this namedtuple. Let's avoid building namedtuples for arguments it just obscures what a function requires and it's general arity complexity.

paz/models/foundation/sonic/pretrained.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+def load_pretrained_weights(encoder, decoder):
+    encoder.load_weights(fetch_asset("sonic_encoder.weights.h5"))
+    decoder.load_weights(fetch_asset("sonic_decoder.weights.h5"))
+
+
+def fetch_asset(filename):
+    return get_file(filename, SONIC_ASSETS_URL + filename,
+                     cache_subdir=SONIC_CACHE_SUBDIR)
```
oarriaga: Let's split into two lines that fit under 80 columns.

paz/models/foundation/sonic/pretrained.py
```diff
@@ -0,0 +1,47 @@
...
+    encoder = build_encoder(layout)
+    decoder = build_decoder(layout)
+    actor = build_actor(layout, encoder, decoder)
+    if weights == "pretrained":
+        load_pretrained_weights(encoder, decoder)
+    return SonicModels(layout, encoder, decoder, actor)
+
+
+def build_layout():
+    obs_config = fetch_asset("observation_config.yaml")
```
oarriaga: Where is this yaml file? Shall we not inline it in order to not depend on additional files?

paz/models/foundation/sonic/layout.py
```diff
@@ -0,0 +1,223 @@
...
+    if not layout.policy_spans:
+        return 0
+    return layout.policy_spans[-1].end
+
+
+def compute_encoder_input_dim(layout):
+    if not layout.encoder_spans:
+        return 0
+    return layout.encoder_spans[-1].end
+
```
oarriaga, line 82: Functions should have a single return and we should do if / else.

oarriaga: This single return policy should be applied throughout the whole PR and specially in this file that seems it makes it harder to track what is being returned explicitly.

oarriaga:
Addressed in e098bb36:

- **`build_input` helper**: inlined at its 3 call sites (`build_actor`/`build_encoder`/`build_decoder`) — agreed, `keras.Input(...)` didn't need the wrapper.
- **`FSQArgs` namedtuple**: removed. This was me over-applying "prefer namedtuples over dicts" during my earlier code-review pass and missing that CLAUDE.md already says *"Pass function parameters explicitly. Do not bundle a function's inputs into an Args namedtuple."* `compute_release_fsq` is back to 7 explicit positional params (matching the original upstream signature); the def now wraps onto 2 lines since there's no way around that many named parameters without either a namedtuple (rejected) or a hanging def — went with the def wrap as the lesser issue.
- **`observation_config.yaml` dependency**: checked the file (80 lines, 2.3KB, pure schema — no trained weights). `SONIC()` now uses a hardcoded `RELEASE_OBSERVATION_CONFIG` dict literal in `pretrained.py` instead of downloading+parsing YAML at runtime; verified its resolved dims still match the real release exactly (1762/994/930/64, 3 modes). `conversion.py`'s asset bundling no longer includes that file either. The general YAML-parsing path (`load_release_observation_layout`) stays in `layout.py` for the conversion CLI, now split into `load_yaml` + `build_observation_layout(config)` so `pretrained.py` can reuse the latter directly.
- **Single return / if-else, throughout**: converted every multi-return function across `layout.py` and `model.py` to a single return via if/else (`compute_policy_input_dim`, `compute_encoder_input_dim`, `compute_temporal_frames`, `compute_feature_groups` in both files, `resolve_observation_dim`, `compute_mode_input`, `compute_flat_mode_input`). Also strengthened CLAUDE.md/AGENTS.md's existing single-return bullet from a soft "prefer... when it improves readability" to a firmer default, since the soft wording apparently wasn't enough to stop these from creeping in in the first place.
- **80-column line splits** on the two spots you flagged: already within 80 after the `build_input` inlining changed their shape; double-checked with an exact column count.

Full suite: `pytest paz/models/foundation/sonic/ examples/sonic/` -> 28 passed, 2 skipped (opt-in real-weight parity, still passing locally against the real release ONNX files with the hardcoded config's dims verified to match exactly).

paz/models/foundation/sonic/conversion.py
```diff
@@ -0,0 +1,120 @@
...
+    actor = build_actor(layout, encoder, decoder)
+    return encoder, decoder, actor
+
+
+def check_input_dim(model, expected_dim, name):
+    input_dim = model.input_shape[-1]
+    if input_dim != expected_dim:
+        raise ValueError(
+            f"{name} input mismatch: expected {expected_dim}, got "
+            f"{input_dim}")
```
oarriaga: Don't split into multiple lines. Either reduce the message size ort split into multiple messages that get concatenated.

paz/models/foundation/sonic/conversion.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+def load_onnx_model(model_path):
+    return onnx.load(str(model_path))
+
+
+def load_onnx_initializers(model):
+    return {tensor.name: numpy_helper.to_array(tensor)
+            for tensor in model.graph.initializer}
```
oarriaga: Do a for loop here. If this type of for loops don't fit in a single line then just use a normal for loop convention.

paz/models/foundation/sonic/conversion.py
```diff
@@ -0,0 +1,120 @@
...
+def apply_encoder_weights(encoder, encoder_onnx):
+    initializers = load_onnx_initializers(encoder_onnx)
+    for name, value in initializers.items():
+        match = _ENCODER_BRANCH_PATTERN.match(name)
+        if match is None:
+            continue
+        branch, layer = match.group("branch"), match.group("layer")
+        layer_name = f"{branch}_module_{layer}"
+        set_dense_weight(encoder.get_layer(layer_name), match.group("kind"),
+                          value)
```
oarriaga: split function into multiple lines such that all functions fits under 80 columns

paz/models/foundation/sonic/conversion.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+def apply_decoder_weights(decoder, decoder_onnx):
+    initializers = load_onnx_initializers(decoder_onnx)
+    kernel_names = find_decoder_kernel_names(decoder_onnx)
+    for layer_id, kernel_name in kernel_names.items():
+        layer = decoder.get_layer(f"g1_dyn_module_{layer_id}")
+        bias_name = f"module.decoders.g1_dyn.module.{layer_id}.bias"
+        layer.set_weights(
+            [initializers[kernel_name], initializers[bias_name]])
```
oarriaga: split function into multiple lines such that all functions fits under 80 columns

paz/models/foundation/sonic/model.py
```diff
@@ -0,0 +1,169 @@
...
+from paz.models.foundation.sonic.layout import compute_decoder_input_dim
+from paz.models.foundation.sonic.layout import compute_encoder_input_dim
+from paz.models.foundation.sonic.layout import compute_mode_scalar_index
+from paz.models.foundation.sonic.layout import compute_policy_tail_dim
+
+
+def build_actor(layout, encoder, decoder):
+    encoder_dim = compute_encoder_input_dim(layout)
+    encoder_obs = keras.Input(
+        shape=(encoder_dim,), dtype="float32", name="encoder_obs")
```
oarriaga: Do you need to explicitly use the shape=(encoder_dim,) maybe just call it without saying "shape" since I think shape is the first argument. That  way hopefully it all fits in a single line.

paz/models/foundation/sonic/model.py
```diff
@@ -0,0 +1,169 @@
...
+    tail_dim = compute_policy_tail_dim(layout)
+    policy_obs = keras.Input(
+        shape=(tail_dim,), dtype="float32", name="policy_obs_tail")
```
oarriaga: Same as the other input layer.

paz/models/foundation/sonic/model.py
```diff
@@ -0,0 +1,169 @@
...
+    decoder_input = compute_cat(encoded_tokens, policy_obs)
+    action = decoder(decoder_input)
+    inputs = {"encoder_obs": encoder_obs, "policy_obs_tail": policy_obs}
+    return keras.Model(inputs, action, name="sonic_actor")
+
+
+def build_encoder(layout):
+    encoder_dim = compute_encoder_input_dim(layout)
+    encoder_input = keras.Input(
+        shape=(encoder_dim,), dtype="float32", name="obs_dict")
```
oarriaga: Same as the other input layer.

paz/models/foundation/sonic/model.py
```diff
@@ -0,0 +1,169 @@
...
+    mode_id = compute_mode_id(encoder_input, compute_mode_scalar_index(layout))
+    branch_tokens = compute_branch_tokens_list(encoder_input, layout)
+    selected_tokens = compute_selected_tokens(
+        branch_tokens, mode_id, len(layout.encoder_modes))
```
oarriaga: Split computation into multiple lines or pack into an args varible and then unpack.

paz/models/foundation/sonic/model.py
```diff
@@ -0,0 +1,169 @@
...
+    return branch_tokens
+
+
+def compute_branch_tokens(encoder_input, mode_layout, token_dim):
+    mode_input = compute_mode_input(encoder_input, mode_layout)
+    hidden_dims = (2048, 1024, 512, 512)
+    args = mode_input, hidden_dims, token_dim, mode_layout.name
+    branch_latent = compute_dense_stack(*args)
+    return compute_release_fsq(
+        branch_latent, 2, 32, 0.032237, 15.515501, 0.5, 16.0)
```
oarriaga: maybe do another args pack and then unpack as above.

paz/models/foundation/sonic/model.py
```diff
@@ -0,0 +1,169 @@
...
+
+def compute_decoder_features(decoder_input, token_dim, tail_dim):
+    token_state = decoder_input[:, :token_dim]
+    policy_tail = decoder_input[:, -tail_dim:]
+    return compute_cat(token_state, policy_tail)
+
+
+def compute_dense_stack(inputs, hidden_dims, output_dim, prefix):
+    x = inputs
+    layer_index = 0
```
oarriaga: you can put these two definitions in a single line.

paz/models/foundation/sonic/model.py
```diff
@@ -0,0 +1,169 @@
...
+        layer_index = layer_index + 2
+    return layers.Dense(output_dim, name=f"{prefix}_module_{layer_index}")(x)
+
+
+def compute_mode_input(encoder_input, mode_layout):
+    if not mode_layout.feature_spans:
+        raise ValueError(f"Encoder mode {mode_layout.name} has no spans")
+    if mode_layout.temporal_frames is None:
+        mode_input = compute_flat_mode_input(
+            encoder_input, mode_layout.feature_spans)
```
oarriaga: split into two lines to avoid breaking them.

paz/models/foundation/sonic/pretrained.py
```diff
@@ -0,0 +1,117 @@
...
+
+
+def load_pretrained_weights(encoder, decoder):
+    encoder.load_weights(fetch_asset("sonic_encoder.weights.h5"))
+    decoder.load_weights(fetch_asset("sonic_decoder.weights.h5"))
+
+
+def fetch_asset(filename):
+    return get_file(filename, SONIC_ASSETS_URL + filename,
+                     cache_subdir=SONIC_CACHE_SUBDIR)
```
oarriaga: Pack into args so that it all fits in a single line or simply do the computation of the string concat a line before and just then put it there to avoid breaking the parenthesis.

oarriaga:
Addressed in d140aff0. Went beyond the 12 flagged spots and swept every file in the package for the same pattern, since it kept recurring:

- **`keras.Input`**: dropped the `shape=` keyword (it's the first positional arg) and the redundant explicit `dtype="float32"` (keras already defaults to float32) — verified against the real release weights that this didn't change numerics.
- **`conversion.py`**: shortened the dim-mismatch message to one line; `load_onnx_initializers` is a real for loop now, not a wrapped dict comprehension; `apply_encoder_weights`/`apply_decoder_weights`/`main()` all precompute their call's pieces (layer, kind, weights, paths) into named locals one line earlier instead of wrapping the call itself.
- **`model.py`**: `compute_selected_tokens`/`compute_release_fsq` calls now pack args into a tuple and unpack with `*args` (matching the pattern this file already used elsewhere); `compute_dense_stack`'s two init lines combined into one; `compute_mode_input` precomputes `mode_layout.feature_spans` before the call.
- **`layout.py`**: same treatment throughout `build_observation_layout`/`load_encoder_mode`; also renamed `encoder_span_map` -> `span_map` so `compute_feature_groups`' call fits on one line without wrapping.
- **`pretrained.py`**: `fetch_asset` precomputes the URL before calling `get_file`.

Also strengthened CLAUDE.md/AGENTS.md's existing hanging-parenthesis rule with the concrete technique that actually resolved all of these (precompute into a named local, or pack args + `*args`, rather than just "extract a helper" which doesn't fit cases like a single `keras.Input(...)` call or a raise message).

Full suite: `pytest paz/models/foundation/sonic/ examples/sonic/` -> 28 passed, 2 skipped (opt-in real-weight parity, still passing locally against the real release ONNX files).

oarriaga:
Pushed weight hosting in be5c906c.

- Checked GR00T-WholeBodyControl's LICENSE: the released model weights are under the **NVIDIA Open Model License** (open, permits redistribution with attribution). Minimal compliance: (a) include a copy of the Agreement, (b) include the attribution notice. Both are now release assets (`sonic_LICENSE.txt`, `sonic_NOTICE.txt`), matching the same License/NOTICE pattern already used for Gemma4's v0.26 release.
- Uploaded to `oarriaga/altamira-data` as **v0.28** (v0.27 was already taken by a different, unrelated release — Depth Anything 3).
- `SONIC(weights="pretrained")` now actually downloads and works: verified end-to-end against the real ONNX outputs (encoder exact match, decoder ~1e-5), same tolerances as the direct-ONNX path, since the hosted weights are byte-derived from it.
- Added `pretrained_test.py`, a real download test matching the existing `detnet`/`iknet` `weights_test.py` convention.
- Re-ran the MuJoCo examples per your request: `mujoco_demo.py --headless` in both g1 and teleop mode, plus the full `simulation_test.py` suite — all still pass.

Full suite: `pytest paz/models/foundation/sonic/ examples/sonic/` -> 29 passed, 2 skipped.

PR #421, opened 2026-07-30
Gemma4 gradient checkpointing and a Florence-2 encoder padding mask

oarriaga:
Two small additions the bulkington VLA work needs. Both default to off, so no
existing path changes.

**Gemma4 gradient checkpointing.** A long multi-image sequence has to keep every
decoder layer's activations for the backward pass, which does not fit beside a
vision tower being tuned. `remat_layers` wraps each layer in `keras.remat` so
they are recomputed instead. Defaults to `False`, and toggling it leaves the
forward bit-for-bit unchanged.

**Florence-2 encoder padding mask.** `build_encoder` already threaded a mask to
`masked_attend`, but `model.build` passed `None`, so a batched caller that pads
its prompts had the padding attended by self attention. Adds a `pad_id`
parameter that derives the key padding mask from the token ids, keeping image
tokens always valid. Deriving it rather than accepting it as an input means a
caller cannot send a mask that disagrees with its tokens.

`pad_id` defaults to `None` so the reference path is untouched: it sends one
exact-length prompt and attends everything, which is what
`test_encoder_hidden_states_match_torch_fixture` compares against. I could not
run that fixture (`FLORENCE2_WEIGHTS`/`FLORENCE2_FIXTURES` are not on this
machine), which is exactly why masking is opt-in rather than default.

Worth a reviewer's eye: whether upstream FLOWER masked padding during its own
training. If it did, masking matches pretraining. If it fed padded batches with
an all-ones mask, masking is a small distribution shift from what the pretrained
encoder saw.

**Verification.** `florence2_test.py` + `flower_test.py`: 14 passed, 2 skipped
(both fixture tests, missing env vars). Three new tests cover the mask: the
default still lets padding reach the real tokens, `pad_id` hides it, and `pad_id`
demonstrably changes the result so the mask cannot silently no-op.

**Revised after review.** This originally also changed `multimodal.py` to accept
precomputed `image_embeddings` and a 4-D frame stack. That was unnecessary:
`Gemma4MultimodalBackbone.call` already routes every image through
`encode_images`, so a subclass in the consuming project covers both cases.
`multimodal.py` is now untouched and the caching lives in bulkington. Force-pushed
to drop it.

### DIFF
--- a/paz/models/foundation/florence2/florence2_test.py
+++ b/paz/models/foundation/florence2/florence2_test.py
@@ -67,3 +67,41 @@ def test_encoder_hidden_states_match_torch_fixture():
     hidden = model.predict([static, gripper, flow_and_text], verbose=0)
     reference = fixtures["encoder_hidden_states"]
     assert np.abs(hidden - reference).max() < 1e-4
+
+
+def test_default_attends_padding():
+    # The default has to behave like the reference, which attends every token,
+    # so appending padding must reach the real tokens as well.
+    model = build(**TINY)
+    images = np.random.default_rng(2).normal(size=(1, 32, 32, 3))
+    images = images.astype("float32")
+    short = np.array([[5, 6, 7]], "int32")
+    padded = np.array([[5, 6, 7, 1, 1]], "int32")
+    kept = model.predict([images, images, short], verbose=0)
+    grown = model.predict([images, images, padded], verbose=0)
+    assert np.abs(kept - grown[:, :kept.shape[1]]).max() > 1e-6
+
+
+def test_pad_id_hides_padding_from_real_tokens():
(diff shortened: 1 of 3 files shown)


oarriaga:
Ordering note: bulkington#24 has already merged and its `main` still calls the `read_images` this PR no longer adds. bulkington#26 replaces that with a subclass on the consumer side. **Merge bulkington#26 before this PR** so the two repositories never disagree.

oarriaga:
Superseded by #422, which carries only the Florence-2 padding mask. That is the piece bulkington `main` needs to build the FLOWER policy. The Gemma4 gradient-checkpointing commit is dropped: only FLOWER is being served, so nothing Gemma4-side is being landed for now. The Gemma4 repair is handed to a follow-up agent.

PR #426, opened 2026-08-04
refactor the whole dino_v2_object_detection

mohamedamrfaroukebrahim:
Refactor the RF-DETR / LW-DETR object detector built on the DINOv2 windowed ViT.


oarriaga:
Thanks @mohamedamrfaroukebrahim — the architecture here, the weight-porting map and the training code are what #427 and #428 are built on.

Plan agreed with the maintainer: this PR merges first so your commit stays in `paz-jax` history. #427 is then rebased on top and its commit removes the vendored `dino_v2_object_detection` tree, replacing the inference path with a version built from paz primitives (1218 lines, verified against `rfdetr` 1.9.1 on all five published COCO checkpoints). #428 refactors the fine-tuning half on top of that.

Two things from this PR that turned out to matter, both now fixed downstream:

- `positional_encoding_size` was accepted but unused, so `RFDETRBase` could not load its own weights — its checkpoint stores DINOv2's 37x37 position grid while the model runs at 40x40. #427 resamples the grid once at weight-port time, which is what the reference's `export()` does.
- The reinitialized class head needs the focal prior bias (`-log(99)`); without it the loss opens near 2141 instead of 1.9.

PR #427, opened 2026-08-06
Add the RF-DETR detector on a windowed DINOv2 backbone

oarriaga:
Refactors #426 down to the inference path of the RF-DETR / LW-DETR detector,
built from paz primitives instead of a vendored copy of the upstream
repository. 1218 added lines against #426's 21646.

## Credit

The architecture, the weight-porting map and the training code this builds on
come from @mohamedamrfaroukebrahim's #426. That PR lands first so its commit
stays in `paz-jax` history; this branch is then rebased on top and its commit
removes the vendored `dino_v2_object_detection` tree, keeping the reviewable
diff to the refactored files. Fine-tuning is refactored from #426's training
code in a follow-up.

## What lands

Reusable primitives in `paz.models.transformers`:

- `windowing` — ViT token window partition / merge
- `deformable` — multi-scale deformable attention
- `embeddings.sine` — DETR normalized-coordinate sine embedding
- `feedforward.relu`
- plus `paz.image.sample_bilinear` (a `grid_sample` equivalent)

The canonical DINOv2 gains a windowed feature backbone,
`DINOv2SmallWindowedFeatures`, which reuses the existing blocks and fused-QKV
attention rather than the `dinov2_legacy` windowed ViT that #426 depended on.
`blocks.build_global` unfolds the windows for the blocks that must attend
image wide.

`paz.models.detection.rf_detr` adds the C2f projector, the two-stage
deformable decoder and `RFDETRNano/Small/Medium/Base/Large`. Each is one flat
functional graph from pixels to `(logits, boxes)` — no layer container
re-applied outside a `Model`, no `SimpleNamespace` facade, no masks or
`NestedTensor` (square-resize inference has no padding, so the reference's
valid-ratio path is the identity).

`paz.applications` adds `DetectRFDETR*` with a jitted model call and a jitted
top-k postprocess, following the `SSD` closure pattern.

`port_weights.py` maps the published PyTorch checkpoints on by name. It also
resamples the stored position grid once at port time, which is what the
reference's `export()` does — that is why `RFDETRBase` works at all, since its
checkpoint stores DINOv2's 37x37 grid while the model runs on 40x40.

## Scope

Inference only. Left out of this PR: the training stack (losses, Hungarian
matcher, COCO loaders, engine, LoRA, EMA, `coco_eval`), the segmentation
variants, and the XLarge/2XLarge Objects365 variants. Only query group zero is
built, which is what the reference uses at inference; group-DETR duplicates the
query tables and first-stage heads for training only.

## Verification

Against `rfdetr` 1.9.1 with the published COCO checkpoints, on CPU. Max
absolute difference on random inputs:

| variant | logits | boxes |
| --- | --- | --- |
| nano | 9.0e-05 | 7.7e-05 |
| small | 1.7e-04 | 1.1e-04 |
| medium | 8.9e-04 | 1.4e-04 |
| base | 2.9e-04 | 9.4e-04 |
| large | 2.3e-04 | 1.5e-04 |

Deformable attention matches the reference core at 1.2e-07 on a two-level,
four-head, three-point case. On COCO val 000000039769 the application
reproduces the reference detections exactly — same four boxes and labels,
scores within 1e-03 (`cat 0.962`, `remote 0.913`, `remote 0.910`, `cat 0.888`).
Jit holds: after one compile, 480x640, 300x500 and 720x1280 inputs all run in
0.088 s with no recompilation, because preprocessing resizes to the fixed
square input.

`parity_test.py` reproduces the table; it is skipped unless
`RFDETR_CHECKPOINTS` points at a directory with the official `.pth` files, and
needs `torch` plus `rfdetr`. It passes 5/5 here.

Suites: `paz/models/transformers` 69 passed, `paz/models/foundation/dinov2` 17
passed 1 skipped, `paz/models/detection/rf_detr` 8 passed 5 skipped,
`paz/backend/image_test.py` 9 passed. `examples/rf_detr/detect.py` smoke-runs
on CPU. The `efficientdet` failures and the `single_shot_detector` collection
errors reproduce identically on `paz-jax`, so they are pre-existing.

## Follow-up needed

`pretrained.py` points at `altamira-data` v0.35, which does not exist yet. The
five converted `.weights.h5` files (103-116 MB each) are built and verified
locally; they need uploading before `DetectRFDETR*` works without a local
weights path.


paz/backend/image.py
```diff
@@ -114,6 +114,60 @@ def bilinear_sample_positions(source, target, ops):
...
 
+def sample_bilinear(images, positions):
+    """Samples images at normalized (x, y) positions inside ``[-1, 1]``.
+
+    Matches ``torch.grid_sample(align_corners=False, padding_mode="zeros")``:
+    corners outside the image contribute zero. Images are channels-last
+    ``(batch, height, width, channels)`` and positions are ``(batch, ..., 2)``.
+    Uses ``keras.ops`` so it traces inside functional models.
+    """
+    from keras import ops
```
oarriaga: This should not be inside a function.

paz/backend/image.py
```diff
@@ -114,6 +114,60 @@ def bilinear_sample_positions(source, target, ops):
...
+    points = int(np.prod(positions.shape[1:-1]))
+    flat = ops.reshape(positions, (-1, points, 2))
+    column = to_pixel_position(flat[..., 0], images.shape[2])
+    row = to_pixel_position(flat[..., 1], images.shape[1])
+    sampled = accumulate_corners(images, column, row, ops)
+    shape = (-1, *positions.shape[1:-1], images.shape[-1])
+    return ops.reshape(sampled, shape)
+
+
+def to_pixel_position(normalized, extent):
```
oarriaga: What is normalized? It's not clear from the variable names what they are.

paz/backend/image.py
```diff
@@ -114,6 +114,60 @@ def bilinear_sample_positions(source, target, ops):
...
+def accumulate_corners(images, column, row, ops):
+    left, top = ops.floor(column), ops.floor(row)
+    corners = (left, top), (left + 1, top), (left, top + 1), (left + 1, top + 1)
+    weights = build_corner_weights(column - left, row - top)
+    sampled = 0.0
+    for corner, weight in zip(corners, weights):
+        inside = compute_inside_image(images, *corner, ops)
+        pixels = take_pixels(images, *corner, ops)
+        sampled = sampled + pixels * ops.expand_dims(weight * inside, -1)
+    return sampled
```
oarriaga: Why does an image accumulate corners? It's unclear what the function paz.image.accumulate_corners is doing. Maybe the name should be more clear.

paz/backend/image.py
```diff
@@ -114,6 +114,60 @@ def bilinear_sample_positions(source, target, ops):
...
+
+
+def build_corner_weights(weight_x, weight_y):
+    left, top = 1.0 - weight_x, 1.0 - weight_y
+    return left * top, weight_x * top, left * weight_y, weight_x * weight_y
+
+
+def compute_inside_image(images, corner_x, corner_y, ops):
```
oarriaga: It's unclear what this function does.

oarriaga: So the name of the function should be different.

paz/models/transformers/embeddings/sine.py
```diff
@@ -0,0 +1,29 @@
...
+    scaled = ops.expand_dims(coordinates * 2 * math.pi, -1) / frequencies
+    sines, cosines = ops.sin(scaled[..., 0::2]), ops.cos(scaled[..., 1::2])
+    interleaved = ops.stack([sines, cosines], axis=-1)
+    shape = (-1, coordinates.shape[1], frequencies.shape[0])
+    return ops.reshape(interleaved, shape)
+
+
+def build_frequencies(dim, dtype):
+    steps = ops.arange(dim, dtype=dtype)
+    return 10000.0 ** (2 * ops.floor(steps / 2) / dim)
```
oarriaga: Wrap the constant into variables such that a reader would understand more precisely why this constants are hardcoded.

oarriaga:
Addressed in da0fe76a.

**Your six comments**

| Comment | Change |
| --- | --- |
| `image.py:125` import inside a function | `from keras import ops` moved to module scope, matching `paz.backend.audio` and `paz.backend.file`; the `ops` parameter I was threading through the helpers is gone. Also removed the now-redundant local import in `resize_bilinear_align_corners`. |
| `:135` unclear what `normalized` is | `to_pixel_position(normalized, extent)` → `to_pixel_coordinate(coordinate, num_pixels)`, with a docstring saying it rescales one axis from `[-1, 1]` into pixel units and that the result is fractional. |
| `:148` why does an image accumulate corners | → `interpolate_pixels(images, column, row)`: "blends the four pixels surrounding each fractional column and row". The corner tuple is now built by `build_neighbors`. |
| `:156` unclear + name should differ (x2) | `compute_inside_image` → `compute_inside_mask`, documented as what makes positions past the border contribute zero rather than repeat the edge. `build_corner_weights` → `compute_neighbor_weights`, documenting the neighbour order and that the weights sum to one. |
| `sine.py:29` wrap the constants | `temperature` and `pair_index` are named locals now. The docstring says the temperature is the ratio between longest and shortest wavelength, fixed by the published weights rather than free to tune, and why consecutive entries share one. |

**Three more from a review pass over the diff**

- `parity_test.py` now pins `jax_default_matmul_precision` itself. The command in its own docstring was failing without `JAX_DEFAULT_MATMUL_PRECISION=highest` — reduced-precision f32 matmuls are the CPU default here and drift ~0.2 in the logits over twelve blocks. It now passes 5/5 as documented.
- `select_proposals` raises at build time when the feature grid has fewer cells than `num_queries`. Previously such a detector built fine and then failed on its first forward pass inside `top_k`.
- `models_test` seeds its weights, so the numerical assertions compare the same graph every run.

**One finding I could not reproduce, so did not act on:** a review pass claimed `test_backbone_jit_matches_eager` fails deterministically at ~2.5e-4 against `atol=1e-4`. Measured across six seeds it is 5.2e-6 to 6.0e-6, and it passes 3/3 without any precision override. I suspect that measurement was of the full model output, which does move by ~5 between batch sizes because ~1e-5 matmul noise reorders the first stage's top-300-of-576 selection — which is exactly why the test targets `projector_norm` instead. I seeded the test rather than loosen a tolerance on a number I could not reproduce.

**Not consolidated, with a reason:** `paz.backend.features.gather_bilinear` does overlap. It is `jax.numpy`-based and unbatched, so it cannot trace on the KerasTensors that `deformable.attend` builds with, and it takes pixel rather than normalized coordinates. Happy to unify if you would rather have one implementation.

#428 is rebased on top and re-verified: 141 passed / 12 skipped, and the VOC fine-tune smoke run still converges (loss 2.07 → val 1.92 on 6 steps).

oarriaga:
Rebuilt on the merged `paz-jax` now that #426 is in. Two commits added:

- `e20bcf95` merges the new `paz-jax` (which brought #426 and #432). No conflicts. I merged rather than rebased so your inline comments stay anchored to `af6cd3b8` and `da0fe76a`.
- `2cfd449e` removes `paz/models/detection/dino_v2_object_detection/` — exactly 21646 deletions, matching what #426 added. Every import of that tree came from inside itself, so nothing else broke.

The PR now reads **+1275 / -21650**, which is the whole point of the exercise: the same detector, built from paz primitives.

**Verified on the merged base**

- 114 passed / 13 skipped across `rf_detr`, `transformers`, `foundation/dinov2`, `backend/image_test`, `backend/boxes_test`, no precision override.
- Parity 5/5 against `rfdetr` 1.9.1 using only the command in `parity_test`'s docstring.
- `examples/rf_detr/detect.py` end to end: `cat 0.962 / remote 0.913 / remote 0.910 / cat 0.888`, unchanged from before the merge.

**Two things I did not do, both your call**

- `dinov2_legacy.models.windowed_vision_transformer` now has no callers — only the deleted tree's tests used it. I left it: the rest of `dinov2_legacy` is still the parity baseline for `paz.models.foundation.dinov2`, and deleting code you deliberately kept isn't mine to decide. Noted in the commit message.
- #426's training half is not replaced. Losses, matcher, COCO loaders, engine, LoRA, EMA and mAP eval are reachable through the merged `0a229962`; #428 refactors the fine-tuning subset only.

#428 is merged up to this and re-verified: 141 passed / 12 skipped. Its base retargets to `paz-jax` automatically once this merges.

PR #428, opened 2026-08-07
Add RF-DETR fine-tuning on VOC

oarriaga:
Stacked on #427 — review that first; the base retargets to `paz-jax` once it
merges. Refactored from the training code in @mohamedamrfaroukebrahim's #426,
scoped to fine-tuning a COCO detector onto a new class set.

## What lands

`paz.losses.detr` — the set-prediction loss: Hungarian matching, IOU-aware
binary cross entropy on the classes, L1 and generalized IOU on the boxes.
Targets are padded with -1 so every tensor keeps a static shape, and the
assignment runs on the host through `jax.pure_callback`, which traces inside
the compiled train step. `paz.boxes` gains `compute_generalized_IOUs` and
`compute_union_areas` beside the existing `compute_IOUs`.

`rf_detr.join_outputs` concatenates boxes and logits so one Keras loss sees
both. `rf_detr.reset_class_heads` biases a reinitialized class head towards a
0.01 prior.

`examples/rf_detr/{generator,fine_tune}.py` follow the existing SSD
`fine_tune.py` convention: COCO weights load with `skip_mismatch=True` so only
the two class heads restart, then `model.compile` and `model.fit`.

`select_detections` no longer drops logit column zero. That column is empty
only because COCO identifiers start at one, so the COCO application carries a
91 entry name list instead, and a model fine-tuned on 20 VOC classes works
through the same postprocess. COCO detections are unchanged.

## Two bugs worth calling out

**`jit_compile` must stay on.** I first assumed Keras could not place the
Hungarian callback inside a compiled train step. It can. Eager ran over eleven
minutes without finishing three steps; jitted it runs at 0.55 s/step.

**A reinitialized class head needs the focal prior bias.** With Keras' default
zero bias the head starts at p≈0.5 over every query and class, and the loss
opens at ~2141. Upstream sets the bias to `-log(99)`; with that it opens at
~1.9.

## Verification

VOC2007, nano, batch 4, frozen backbone, 300 steps per epoch, CPU:

| epoch | loss | classification | GIoU | L1 | val_loss |
| --- | --- | --- | --- | --- | --- |
| 0 | 1.556 | 1.017 | 0.150 | 0.048 | 1.401 |
| 1 | 1.371 | 0.805 | 0.149 | 0.054 | 1.299 |
| 2 | 1.278 | 0.730 | 0.149 | 0.050 | 1.247 |
| 3 | 1.257 | 0.719 | 0.147 | 0.049 | 1.362 |

Validation bottoms at epoch 2 and ticks up at epoch 3, as expected after only
~4800 images. On VOC2007 test the epoch-3 checkpoint recovers the annotated
classes:

```
000001.jpg  truth [dog, person]                     predicted person:0.71 dog:0.65
000002.jpg  truth [train]                           predicted train:0.82
000004.jpg  truth [car]                             predicted car:0.83 ... car:0.31
000006.jpg  truth [chair, diningtable, pottedplant] predicted chair:0.82 diningtable:0.43 pottedplant:0.33
```

Suites: `paz/losses` 22 passed, `paz/backend/boxes_test.py` 15 passed 7
skipped, `paz/models/detection/rf_detr` 8 passed 5 skipped. The seven
`boxes_test` pyflakes hits are pre-existing.

## Scope left out

Each costs accuracy against upstream and is worth a follow-up: auxiliary and
encoder losses (they need intermediate decoder outputs exposed), group-DETR's
thirteen query groups, and every augmentation but the horizontal flip. paz's
colour helpers assume 0-255 images and its scale-jitter helpers move boxes out
of normalized coordinates, so neither composes with this pipeline as written.


oarriaga:
Retargeted to `paz-jax` and merged up to it now that #427 is in (`b7ce730b`, no conflicts). The diff is unchanged at **+594 / -9 across 11 files** — it was always additive on top of #427.

**Verified on the merged base**

- 158 passed / 13 skipped across `paz/losses`, `rf_detr`, `transformers`, `foundation/dinov2`, `backend/boxes_test`, `backend/image_test`, no precision override.
- `examples/rf_detr/fine_tune.py` on VOC2007 still converges: loss 2.07 → val 1.92 over 6 steps at batch 2 with a frozen backbone.

The longer run recorded in the PR body still stands (4 epochs, train 1.556 → 1.257, and the epoch-3 checkpoint recovering dog/person, train, car, and chair+diningtable+pottedplant on VOC2007 test).

One thing to weigh before merging: `DetectRFDETR*` and this script both start from the COCO weights, and `pretrained.py` points at `altamira-data` v0.35, which does not exist yet. Merging is safe, but neither the application nor fine-tuning runs on a fresh clone until those five `.weights.h5` files are uploaded.

PR #435, opened 2026-08-22
Move the multi-view mesh fitting example to examples/multi_view

oarriaga:
Moves the cow fitting script to examples/multi_view/optimize_mesh.py and rewrites it in the current example style: argparse inlined in the main guard, the Args namedtuples replaced with explicit parameters and module constants, and paz.draw.mosaic / compute_triangle_normals reused instead of local copies. The renderer and loss knobs are now constants; the remaining flags are mesh_dir, output_dir, max_steps and seed.

Verified on GPU: a 20 step smoke run reproduces the previous initial metrics exactly (initial loss 0.27668), and a full 2000 step run reaches best loss 0.0079 at step 1707, matching the last run of the old script.


examples/multi_view/optimize_mesh.py
```diff
@@ -0,0 +1,422 @@
...
+from paz.graphics.mesh import compute_triangle_normals
+from paz.graphics.mesh import count_binned_faces, tile_render_binned_soft_mask
+from paz.graphics.types import Material, Pattern, PointLight
+from paz.optimization.history import trim_trace
+
+IMAGE_SHAPE = (128, 128)
+Y_FOV = jp.pi / 3.0
+NUM_VIEWS = 20
+MASK_SIGMA = 1e-4
+BINS = BinArgs(16, 4608)
```
oarriaga: These should be parameters of argparse.

examples/multi_view/optimize_mesh.py
```diff
@@ -0,0 +1,422 @@
...
+
+Parameters = namedtuple("Parameters", "offsets vertex_colors step_arg")
+Views = namedtuple("Views", "images masks poses")
+LossTerms = namedtuple("LossTerms", "rgb silhouette edge normal laplacian")
+WEIGHTS = LossTerms(1.0, 1.0, 1.0, 0.01, 1.0)
```
oarriaga: No need to create so many abstractions. Also weights should be names of argparse.

PR #436, opened 2026-08-29
Finish robust walking policy training in rl_gym

oarriaga:
Completes the from-scratch robust walking policy for the G1 in examples/rl_gym: wires per-environment physics randomization into training, adds principled simulator-divergence handling (non-finite or unphysically fast states terminate without bootstrap and are masked out of the loss, with a divergence count in the training log), multi-GPU training with one process per GPU and globally averaged gradients (matching the reference DDP setup), Keras-based checkpointing with resume, a playback script, a README, and co-located numerical tests.

Bug fixes found along the way: randomize.physics_model passed the joint count as joint_dry_friction's minval (~29 N-m of dry friction on every joint once randomization was wired in); njmax was scaled by num_envs although warp budgets it per environment, a quadratic memory blow-up that made 4096 envs impossible; networks.PPO ignored its num_actions argument.

Deliberate divergences from the rsl_rl reference, documented in the README: per-epoch experience reshuffling, log-parameterized action noise, and the in-graph divergence handling.

Verification: 49 pytest cases pass (JAX_PLATFORMS=cpu pytest examples/rl_gym); single-GPU and two-GPU smoke runs at 512 to 4096 envs train with finite metrics (~45k steps/s single GPU at 4096 envs); checkpoint resume continues at the saved iteration with restored learning rate and speed curriculum; play.py writes a rollout video. /code-review at medium effort produced 10 findings; 9 were fixed and re-verified, one (a claimed njmax jax-backend regression) was refuted since make_data ignores those budgets for impl='jax'.


oarriaga:
Follow-up from the full 8192-env replication run: the divergence bound drops to 100 rad/s and diverged samples are excluded from advantage normalization (finite sub-threshold explosions were flattening every other advantage, letting the entropy bonus inflate the noise); the action noise is reverted to rsl_rl's direct parameterization with a positive floor (the log parameterization shrank noise to 0.61 and stalled exploration in a standing reward-hack); rendering fixes make the collision-only robot and its translation visible in play.py videos. 51 tests pass; a fresh 30000-iteration replication is running.

examples/rl_gym/simulation/common.py
```diff
@@ -157,6 +159,29 @@ def apply_scheduled_push(key, physics_state, episode_step, push_step):
...
+def sanitize_diverged(physics_state, dynamics, velocity_bound=100.0):
+    # a non-finite or unphysically fast state is a solver failure, not an
+    # outcome: replace it so no poison reaches the observation history, and
+    # report the divergence; the bound sits several times above the joint
+    # velocity limits, so it also caps how large any reward term can get
+    # before an explosion is caught
+    finite_qpos = jp.all(jp.isfinite(physics_state.qpos))
+    speed = jp.max(jp.abs(physics_state.qvel))
+    stable_qvel = jp.isfinite(speed) & (speed < velocity_bound)
+    diverged = ~(finite_qpos & stable_qvel)
```
oarriaga: Let's use always use the logical operators so jp.logical_not it makes it more explict and easy to read.

